PR #1, opened 2019-11-22
Masks example notebook

takluyver:
It should now be easy to make masks using the geometry, and you don't need to assemble images to use the mask - e.g. if you want to sum intensity in a given region.

I've put together an example to highlight this.


tmichela:
One remark on the form: It seems that people use the masks the other way around (mask zones to ignore),
else LGTM

takluyver:
I thought both ways are used in different places? In any case, it's a good
point that we should mention inverting the mask with ~mask.

tmichela:
> I thought both ways are used in different places?

OK, I'm probably biased by the kind of experiment I've seen most

takluyver:
I've added a note about how to get the other kind of mask.

tmichela:
Thansk; LGTM

fangohr:
Thanks - looks great.

PR #4, opened 2019-12-16
Fix compare() method

### LINKED ISSUE #3: Compare arrowheads dwarf the actual geometry
```python
geom.compare(geom2)
```

image

Probably related to the switch from pixel units to metres.

takluyver:
Fixes #3 

This also makes `.compare()` use `.inspect()`, so it draws the same kind of picture. And I moved it onto the base class, so now all detector types support comparing two geometries.

Here's an example image from the new compare method:

image

### DIFF
--- a/docs/geometry.rst
+++ b/docs/geometry.rst
@@ -92,6 +92,8 @@ which this geometry code can position independently.
 
    .. automethod:: inspect
 
+   .. automethod:: compare
+
    .. automethod:: data_coords_to_positions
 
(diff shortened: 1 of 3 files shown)


tmichela:
no support for `z` axis ? :-D 

LGTM

takluyver:
It's a safety concern, if the arrows point out of the screen they could poke the user in the eye. :wink: (someone holding their bruised eye shut)

PR #6, opened 2020-01-29
Jungfrau flex geom

dallanto:
@takluyver  
Pls. have a look into my 2 commits. The first one corresponds to what we discussed on Monday. The second one (current state) adds the loop over modules based on origin positions given.  
This works, but I wonder how serious the warning about the mutable input = list argument is.  
Moreover I tried to pass a list of orientation tuples (1,1) or (-1,-1) etc. The axis flipping is indeed effective, but only on tile level: I am now looking into a re-definition in the corner_x (_y) "equation" in order to flip also the corners, and hence the overall module.

### DIFF
--- a/extra_geom/detectors.py
+++ b/extra_geom/detectors.py
@@ -1349,13 +1349,15 @@ class DSSC_Geometry(DSSC_1MGeometry):
 
 
 class JUNGFRAUGeometry(DetectorGeometryBase):
-    """Detector layout for flexible Jungfrau arrangements based on the JF-500K
-       module, which is an independent detector unit of 2 x 4 ASIC tiles.
+    """Detector layout for flexible Jungfrau arrangements
 
-       The ss/fs assignment is the other way around as for AGIPD:
-       y = ss
-       x = fs
-       shape = (ss, fs) = (y, x)
+     The base JUNGFRAU unit (and rigid group) in combined arrangements is the
+     JF-500K module, which is an independent detector unit of 2 x 4 ASIC tiles.
+
+     The slow-scan/fast-scan assignment is:
+     y = ss
+     x = fs
+     shape = (ss, fs) = (y, x)
     """
     detector_type_name = 'JUNGFRAU'
     pixel_size = 7.5e-5   # 7.5e-5 metres = 75 micrometer = 0.075 mm
@@ -1372,24 +1374,24 @@ class JUNGFRAUGeometry(DetectorGeometryBase):
                               asic_gap=2, unit=pixel_size):
         """Generate a Jungfrau geometry object from module positions
 
-           Parameters
-           ----------
+        Parameters
+        ----------
 
-           offsets - iterable of length n_modules containing a pixel coordinate
-                     tuple (x,y) for each offset to the global origin
+        offsets - iterable of length n_modules containing a pixel coordinate
+                  tuple (x,y) for each offset to the global origin
 
-           orientations - iterable of length n_modules containing a unit-vector
-                     tuple (x,y) for each orientation wrt. the axes
+        orientations - iterable of length n_modules containing a unit-vector
+                  tuple (x,y) for each orientation wrt. the axes
 
-           Orientations default to (1,1) for each module if this optional
-           keyword argument is lacking; if not, the number of elements must
-           match the number of modules as per offsets
+        Orientations default to (1,1) for each module if this optional
+        keyword argument is lacking; if not, the number of elements must
+        match the number of modules as per offsets
 
-           We assume that externally defined offsets as per input relate the
-           bottom, beam-left pixel of the module to the global origin, which is
-           bottom, beam-left of the overall assembly.
-           This is a definition *before* flipping: if flipping is present, we
-           will adjust for that.
+        We assume that externally defined offsets as per input relate the
+        bottom, beam-left pixel of the module to the global origin, which is
+        bottom, beam-left of the overall assembly.
+        This is a definition *before* flipping: if flipping is present, we
+        will adjust for that.
         """
         px_conversion = unit / cls.pixel_size
         # fill orientations with defaults to match number of offsets
@@ -1477,3 +1479,12 @@ class JUNGFRAUGeometry(DetectorGeometryBase):
         ss_slice = slice(tile_ss_offset, tile_ss_offset + cls.frag_ss_pixels)
         fs_slice = slice(tile_fs_offset, tile_fs_offset + cls.frag_fs_pixels)
         return ss_slice, fs_slice
+
+    @classmethod
+    def from_crystfel_geom(cls, filename):
+
+        raise NotImplementedError
+
+    def write_crystfel_geom(self, file_name):
+
+        raise NotImplementedError
--- a/extra_geom/detectors.py
+++ b/extra_geom/detectors.py
@@ -1344,9 +1344,9 @@ class DSSC_Geometry(DSSC_1MGeometry):
         )
 
 
-class JF_500KGeometry(DetectorGeometryBase):
-    """Detector layout for Jungfrau-500K,
-       which is a single panel of 2 x 4 ASIC tiles.
+class Jungfrau_Geometry(DetectorGeometryBase):
+    """Detector layout for flexible Jungfrau arrangements based on the JF-500K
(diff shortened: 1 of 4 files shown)


dallanto:
Update:
with a new commit I added code that seems to do the correct thing as per quick testing, however I make an assumption based on the sparse EuXFEL/Jungfrau geometry information that we have: that origins are defined as offsets independently from the flip yes/no information, unlike for AGIPD where the quadrant positions already account for the static "flipping scheme" of the topology.

dallanto:
@takluyver 
The edits/suggestions after our last discussion have been committed, except for the zipping of offsets and orientations. I need to consider once more: while the code would look neater, I guess, I feel that zipping both together will be less intuitive when reading the zipped iterable name e.g. "x_transform", where directional scale-factor (orientation 1 or -1) and translation (offset) would then account for different tuple indices. But maybe you had a more clever solution in mind anyway ...

extra_geom/detectors.py
```diff
@@ -1342,3 +1342,119 @@ def __init__(self, modules, filename='No file'):
         warnings.warn(
             "DSSC_Geometry has been renamed to DSSC_1MGeometry.", stacklevel=2
         )
+
+
+class Jungfrau_Geometry(DetectorGeometryBase):
```
takluyver: Naming nitpicks: PSI always write JUNGFRAU in block capitals, as it's technically an acronym, even if it's a really awkward one. And for the other detectors, the underscore is within the detector type name, which we don't need here.

So: `JUNGFRAUGeometry`? Or do we ignore their capitalisation and go for the more legible `JungfrauGeometry`?

extra_geom/detectors.py
```diff
@@ -1342,3 +1342,119 @@ def __init__(self, modules, filename='No file'):
...
+    @classmethod
+    def from_module_positions(cls,offsets=((0,0),), orientations=None,
+                              asic_gap=2, unit=pixel_size):
+        """Generate a Jungfrau geometry object from module positions
+
+           Parameters
+           ----------
+
+           offsets - iterable of length n_modules containing a pixel coordinate
+                     tuple (x,y) for each offset to the global origin
```
takluyver: I think we should specify which corner of the module the offset is to. I think the code currently uses it as the position of the lower beam-right corner?

My preference is to specify the position of the (0, 0) pixel, but that does mean that which corner you're measuring to depends on the orientation, which I appreciate is not necessarily the most convenient way of working. So I'm happy to leave the implementation as is, but I would like it to be clarified in the docstring.

I see this is in an internal comment in the loop already, but that's not visible without inspecting the source code. And I prefer to avoid unqualified 'left' and 'right', as I mentioned, because people look at it in both orientations. I hope that terms like 'beam-left' are a clear shorthand for 'left when looking along the beam'.

extra_geom/detectors.py
```diff
@@ -617,6 +618,7 @@ class AGIPD_1MGeometry(DetectorGeometryBase):
     You won't normally instantiate this class directly:
     use one of the constructor class methods to create or load a geometry.
     """
+    detector_type_name = 'AGIPD-1M'
```
tmichela: You use `-` as separator here, but `_` for other classes. Is it intended?

extra_geom/detectors.py
```diff
@@ -1342,3 +1346,131 @@ def __init__(self, modules, filename='No file'):
         warnings.warn(
             "DSSC_Geometry has been renamed to DSSC_1MGeometry.", stacklevel=2
         )
+
+
+class JUNGFRAU_Geometry(DetectorGeometryBase):
```
tmichela: That should be `JUNGFRAUGeometry` to be consistent, right?

extra_geom/detectors.py
takluyver: I think it should be `-` for all, as this is going into a human-readable comment.

extra_geom/detectors.py
takluyver: Yup, I agree. The `_` in the others is standing in for the dash in e.g. 'AGIPD-1M'. We don't need that for JUNGFRAU.

extra_geom/detectors.py
dallanto: JUNGFRAUGeometry.  
Note that I got rid of the redundant "_" as well

extra_geom/detectors.py
dallanto: I shortened the internal comment and put most of its meaning into the docstring, rephrasing a bit (e.g. using beam-left).

extra_geom/detectors.py
dallanto: Allright, "-" in the detector name strings, and "_" for "-" in the class names, where required.

extra_geom/detectors.py
dallanto: Corrected. See comment above.

dallanto:
Pls. don't get confused: because the last commit was minor (no code changes), I squashed with the previous one and rebased.

extra_geom/detectors.py
```diff
@@ -1342,3 +1346,134 @@ def __init__(self, modules, filename='No file'):
         warnings.warn(
             "DSSC_Geometry has been renamed to DSSC_1MGeometry.", stacklevel=2
         )
+
+
+class JUNGFRAUGeometry(DetectorGeometryBase):
+    """Detector layout for flexible Jungfrau arrangements based on the JF-500K
+       module, which is an independent detector unit of 2 x 4 ASIC tiles.
+
+       The ss/fs assignment is the other way around as for AGIPD:
```
takluyver: Nitpick: let's remove the reference to AGIPD - there's no particular reason someone looking at JUNGFRAU geometry should be familiar with AGIPD.

Also, let's spell out slow-scan & fast-scan in the docstring (even if I haven't done this everywhere ;-) ). The abbreviations aren't familiar to everyone, and I know people who prefer to avoid the initials 'ss' even in innocent contexts.

extra_geom/detectors.py
```diff
@@ -1342,3 +1346,134 @@ def __init__(self, modules, filename='No file'):
...
+           orientations - iterable of length n_modules containing a unit-vector
+                     tuple (x,y) for each orientation wrt. the axes
+
+           Orientations default to (1,1) for each module if this optional
+           keyword argument is lacking; if not, the number of elements must
+           match the number of modules as per offsets
+
+           We assume that externally defined offsets as per input relate the
+           bottom, beam-left pixel of the module to the global origin, which is
+           bottom, beam-left of the overall assembly.
```
takluyver: The global origin can be anywhere, but our general convention is to use it as the centre of the geometry - or the point where a line from the sample is normal to the detector surface (the PONI, Point Of Normal Incidence in PyFAI terms).

extra_geom/detectors.py
dallanto: This has been corrected with the last commit

extra_geom/detectors.py
dallanto: Ok. I am aware that the origin is in the center for e.g. AGIPD.  
The way we looked at the 'inspect' scheme (and also assembled data image) for the JUNGFRAU WIP, the origin was bottom-beam-left and so far I kept it like this.  
The difference is that the big detectors have a center by quadrant-design, while for JUNGFRAU this is less obvious. The way the offsets are (currently) defined in the Karabo scene, it seems to suggest strongly a reference to bottom beam-left overall.

In principle one can transform (translate) everything relative to a the nominal pixel center (0,0), which would reflect diffraction geometry better for any radial-symmetric pattern, but for arrangements like 2 x 3 units it will also "split" one module into two coordinate system hemispheres.

dallanto: I'd also consider that the actual origin for a diffraction geometry should be the beam center, which isn't necessarily the nominal detector center (midpoint of assembled pixel image).  
Once the determination of beam center is done - a possible feature that DA can provide in the future, I guess - one can still relate all pixel coordinate to that.

dallanto: Respectively to the PONI ...

takluyver:
Thanks @dallanto ! I'd still like to think about exactly how we handle the origin, detector centre, beam centre, and PONI, because I think it's valuable to have it consistent between classes. But I'll open a separate issue for that.

PR #7, opened 2020-02-06
Correct magic number in 1M geometry code

zhujun98:
If I understood the geometry code correctly, it is a coincidence that LPD has the same number of modules as the number of tiles per module.

There are a few other places using a magic number instead of the class attributes. If you think it make sense, I will correct them as well.

### DIFF
--- a/extra_geom/detectors.py
+++ b/extra_geom/detectors.py
@@ -981,7 +981,7 @@ class LPD_1MGeometry(DetectorGeometryBase):
                 mod_offset = mod_grp['Position'][:2]
 
                 tiles = []
-                for T in range(1, cls.n_modules+1):
+                for T in range(1, cls.n_tiles_per_module+1):
                     corner_pos = np.zeros(3)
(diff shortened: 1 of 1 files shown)


takluyver:
Yup, that looks right, thanks.

Personally, I often find code clearer with some magic numbers in - e.g. (16, 256, 256) seems easier to read than (nmodules, module_px_slow, module_px_fast). What I'd call a constant might not be the name you'd pick, but the number is the same for both of us. But I'm aware that the prevailing opinion is that magic numbers are to be avoided, and there are obviously times when they cause ambiguity. So I won't object if anyone wants to replace magic numbers with names.

zhujun98:
I agree that when it comes to big matrix, using the 'constant' will not cause too much ambiguity. Actually, by "magic number" here, I mean the number in the for loop. For example, in the DSSC geometry there is `for T in range(1, 3)`. It is not so expressive for someone who reads the code for the first time.

Since you have merged this MR, I am not keen to make another MR to do this. Maybe someone can do it conveniently when making some other changes in the future.

Thank you!

PR #13, opened 2020-06-23
Add support for pnCCD detector layout

philsmt:
Hi there,
this PR adds the somewhat simple geometry of the pnCCD detector, currently available to users at the SQS instrument. Please see here for its specification: https://www.xfel.eu/facility/instruments/sqs/instrument/nqs_end_station/index_eng.html

As the detector consists of only two _parts_, I am not sure whether to treat it as two modules with one tile each or one module with two tiles. Thus, I've actually created a version for both, with this branch implementing the former configuration and [pnccd_geom_1m2t](/European-XFEL/EXtra-geom/tree/pnccd_geom_1m2t) the latter. Note that the data for this detector is always read-out in one matrix for the complete frame, i.e. both _parts_ are concatenated. The public methods of `DetectorGeometryBase` are overriden to support this data format.

Personally, I find the 2m1t version expresses the separation more explicit, as it is visible in the data shape and not somewhat hidden behind the tile support. If the separation would ever become more complicated than simple reshaping however, this machinery would definitely be worthwhile to use. In the end, both versions are comparably simple, so I'd like to go with the version that fits the original design of EXtra-geom the most.
As a small side note, the scientists typically refer to it as modules rather than tiles.

EDIT: Please find a sample image for testing at /gpfs/exfel/data/scratch/schmidtp/pnccd_example.h5

PS: Not planning to skip on tests, but I'd really like to first get this in for the upcoming beamtime.

### DIFF
--- a/extra_geom/detectors.py
+++ b/extra_geom/detectors.py
@@ -182,13 +182,13 @@ class DetectorGeometryBase:
         """
         from matplotlib.collections import PatchCollection
         from matplotlib.patches import FancyArrow
-        
+
         coord_scale = 1 / self.pixel_size
         arrow_scale = scale * coord_scale
 
         # Draw this geometry first, using pixel units
         ax = self.inspect()
-        
+
         if len(self.modules) != len(other.modules):
             print("Geometry objects have different numbers of modules!")
         if any(len(mod_a) != len(mod_b) for (mod_a, mod_b) in zip(self.modules, other.modules)):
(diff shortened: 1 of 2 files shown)


takluyver:
This (both versions) looks pretty good.

Some other methods may be relevant for the choice between 2 modules/tiles:

- `get_pixel_positions()` returns an array of shape `expected_data_shape + (3,)`, i.e. (2, 512, 1024, 3) in this variant or (1, 1024, 1024, 3) in the other.
- `data_coords_to_positions()` expects a module number and pixel coordinates within that module.

This really depends how people are likely to use the data, though - if you refer to a feature in the un-assembled image, do you look at y pixel 720, or at lower module y pixel 208?

extra_geom/detectors.py
```diff
@@ -1494,3 +1494,149 @@ def from_crystfel_geom(cls, filename):
...
+
+        Parameters
+        ----------
+
+        top, bottom: array_like of length 3
+          Absolute position (x, y, z) for the first pixel of each
+          module.
+        """
+
+        args = (np.array([0, -cls.pixel_size, 0]),
```
takluyver, line 1586: Just to be sure, have we checked that the data is stored top to bottom? That's the default in imshow, and if the images are symmetric it may not matter much to show them upside down, but we should get it right.

takluyver:
I'm OK with adding this without tests at first. It's fairly straightforward, and there's little risk of breaking any existing code by adding a new class. Obviously it's still worth thinking about, because once it's in use we'll have a responsibility to support it.

extra_geom/detectors.py
philsmt: Right! I assumed the readout to look upstream and be top-down, but I've pinged the instrument for confirmation. In the worst case, we can check this during commissioning. There will also be an engineer from the manufacturer present.

philsmt:
> * `get_pixel_positions()` returns an array of shape `expected_data_shape + (3,)`, i.e. (2, 512, 1024, 3) in this variant or (1, 1024, 1024, 3) in the other.
> * `data_coords_to_positions()` expects a module number and pixel coordinates within that module.
> 
> This really depends how people are likely to use the data, though - if you refer to a feature in the un-assembled image, do you look at y pixel 720, or at lower module y pixel 208?

I'd assume the hardware readout is fairly artifical and was only used as online feedback for its availability (and lack of better tools at the time, which changed now). Given the modules are generally not aligned in x or y, using a common coordinate doesn't seem sensible. I think this makes the case for 2 modules and 1 tile even stronger. Maybe @egorsobolev can comment on this?

egorsobolev:
I would suggest using modules because they can move relative to each other, and they do not have any common readout channels, i.e. they are completely independent.
In addition, I think that several ASICS are actually lithographically etched on the same plate. I am not sure here. But if I'm right, then it makes sense to provide a method for reshaping arrays with respect to them.

egorsobolev:
I checked my old notebooks. Locking on data I would say that image splits on:

takluyver:
Is there a physical gap between those smaller elements? I.e. do we need to position them separately, or do they make a neat cartesian grid of pixels, even if they're somewhat separate in the hardware?

egorsobolev:
Images look monolithic, I think there are no other gaps, except the gap between modules. Thus, no need for separate positioning of smaller elements

And I am also not sure about MUXs because images look striped and it is difficult to combine stripes into something bigger.

takluyver:
OK, so it seems like it's OK to keep it as 2 modules of 1 tile each, and we believe we've got the orientation correct. @philsmt, is there anything else you want to do before we merge this? And do you want a new release to follow, or will you use it from master?

I'll open issues for tests and docs so we don't overlook them. :-)

philsmt:
At the moment, I think that is all we need. I just force-pushed a fix for some comment typos,

The current users will most likely not use it themselves, but I need it for centre optimisation with geoAssembler. I suppose it would be nice to have it in the exfel environment, so it's seamless for users to execute any code using it. But I can also create a special environment for that purpose, if it doesn't warrant a new release.

takluyver:
OK, EXtra-geom 0.10 is out, and installed in `exfel_anaconda3`.

PR #16, opened 2020-07-16
Reduce overhead for assembling images

takluyver:
In two areas:

1. Apply the centre offset (moving the origin from the centre to the corner) when making the snapped geometry, instead of leaving it until we assemble an image. This assumes you make a geometry once and assemble many images with it.
2. `numpy.split()` seems to be surprisingly costly to call - replacing it with list comprehensions and slicing makes `split_tiles()` up to an order of magnitude faster.

This makes most difference for assembling single images, and for detectors with lots of tiles. I see roughly a 35% speedup on a single image for LPD, 25% for AGIPD, and about 10% for DSSC.

### DIFF
--- a/extra_geom/tests/test_geometry_old.py
+++ b/extra_geom/tests/test_geometry_old.py
@@ -16,7 +16,7 @@ quadpos = [(-11.4, -299), (11.5, -8), (-254.5, 16), (-278.5, -275)]
 
 
 def test_inspect():
-    with h5py.File(pjoin(tests_dir, 'lpd_mar_18.h5')) as f:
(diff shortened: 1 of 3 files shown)


extra_geom/tests/test_geometry_old.py
```diff
@@ -16,7 +16,7 @@
 
 
 def test_inspect():
-    with h5py.File(pjoin(tests_dir, 'lpd_mar_18.h5')) as f:
```
takluyver, line 19: This is unrelated, but it was producing warnings in the test, and it's an easy fix.

tmichela:
LGTM

PR #17, opened 2020-07-21
Parallel image assembly

takluyver:
The default is still serial. My tests suggest it's still quicker to assembling a single (AGIPD) image in serial: I guess the overheads of distributing the work outweigh the benefit of copying small amounts of memory in parallel. The benefit comes when you have a stack of images: the parallel code draws level by about 10 MB, and with 1.4 GB we can get more than an order of magnitude speedup.

This is based on threads, not processes, because the workers need to write to a single destination array. So we're relying on numpy releasing the GIL while copying data. In principle it could also be done with processes if you ensured that both the source and the destination arrays were in shared memory, but I haven't tried to tackle that.

I went for passing a threadpool in so you can reuse a threadpool. In practice, it appears to be so quick to create & tear down threads that this doesn't really matter, but I'm inclined to leave it that way anyway. You could also pass a process pool in if you want to experiment with shared memory, though I'm not documenting such possibilities.

### DIFF
--- a/extra_geom/detectors.py
+++ b/extra_geom/detectors.py
@@ -333,7 +333,7 @@ class DetectorGeometryBase:
         return self._snapped().make_output_array(extra_shape=extra_shape,
                                                  dtype=dtype)
 
-    def position_modules_fast(self, data, out=None):
+    def position_modules_fast(self, data, out=None, threadpool=None):
         """Assemble data from this detector according to where the pixels are.
 
         This approximates the geometry to align all pixels to a 2D grid.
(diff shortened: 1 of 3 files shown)


tmichela:
Would it make sens to have a simple threshold to 'activate' the use of threads: Figuring out the number of images it starts to be faster to run the assembling in a threadpool, and use it only above that threshold?

takluyver:
I wondered about that, but I think it's too 'smart'. The threshold probably depends on your hardware, what else is going on (is it already parallelised at some higher level), and on whether or not the data has to be transposed to assemble it. So I don't want to pick a specific threshold to hard code. It's also valuable to have consistency - if you work with a small sample of your data, it still runs exactly the same code.

tmichela:
LGTM

takluyver:
Thanks!

I investigated the performance with LPD & DSSC as well. In each case, parallelisation is an advantage with a plausible amount of data for one train. With LPD, I needed about 80 MB of data (40 float32 frames) before the parallel code was faster. With DSSC, I only needed about 6 MB (3 uint16 frames).

This was somewhat counterintuitive for me: I expected that parallelisation would be most effective where there were many tiles to place (LPD > AGIPD > DSSC), but the reverse appears to be true. I think the explanation is that more tiles means smaller tiles, so a larger fraction of the time per tile is overhead where the GIL is still held. This approach only parallelises the memory copy where numpy releases the GIL, so it's more effective with larger chunks.

If we want more performance, we could also look at using multiprocessing & shared memory (avoiding the GIL), or using Cython or Numba to do more of the logistics without the GIL.

tmichela:
Nice, thanks for the details :+1:

PR #23, opened 2020-08-28
Fix y-axis scale for DSSC data plotted in metres

takluyver:
The `axis_units='m'` option when plotting puts the scale in metres rather than pixels. This was implemented as a simple scale factor, which works where pixels have equal aspect ratio. DSSC pixels don't, and we were scaling by the pixel width, exaggerating the vertical height.

This shouldn't change the images, which we were already squashing to the correct aspect ratio, only the y ticks.


tmichela:
LGTM

PR #24, opened 2020-09-18
Methods to retrieve quadrant positions and write XFEL HDF5 geometry files

### LINKED ISSUE #18: Write XFEL HDF5 geometry files
This came up the other day for DSSC - I was able to create geometry from another source, but then had no way to write it to a file.

Should it return quadrant positions separately (matching the HDF5 geometry files we've got from DET), or save the quadrant positions in the file (more convenient to use, assuming the quadrants aren't moved frequently, and we write code to read them back in)?

### LINKED ISSUE #20: Get quadrant positions from a geometry object
This would be useful for geoAssembler, which currently implements the same thing itself. Quadrant positions should be suitable for re-creating the same geometry object.

Complication: If you're using an XFEL format geometry file, each module & tile has its own offsets from the overall quadrant position. EXtra-geom stores an absolute position for each tile, rather than a hierarchy of offsets. So we need a way to get the quadrant positions given the offsets for a particular module & tile.

This is also related to #18.

takluyver:
More ways of getting geometry out. These are closely linked because the XFEL geometry files store offsets from quadrant positions. So to get quadrant positions compatible with an XFEL file, you need to either create both together, or read the file when creating quadrant positions. This will be used in geoAssembler.

I'm aiming for consistency within detector classes more than betweeen them, so quadrant positions for LPD & DSSC are returned in mm (like XFEL geometry files), but quadrant positions for AGIPD are in pixel units (like CrystFEL geometry files). Not ideal, but I figure that it's more important to easily roundtrip each geometry to/from files & quadrant positions than it is to compare geometries for different detectors.

I started writing an internal abstraction over the XFEL HDF5 geometry format, but the generalisation was adding much more code than it was saving, so I abandoned it for now.

Closes #18
Closes #20

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 language: python
 python:
+  - 3.8
   - 3.7
   - 3.6
-  - 3.5
 
 env:
   global:
--- a/docs/geometry.rst
+++ b/docs/geometry.rst
@@ -41,6 +41,8 @@ but this geometry code works with a position for each tile.
 
    .. automethod:: from_crystfel_geom
 
+   .. automethod:: quad_positions
+
    .. automethod:: write_crystfel_geom
 
    .. automethod:: get_pixel_positions
@@ -81,6 +83,10 @@ which this geometry code can position independently.
    .. automethod:: from_h5_file_and_quad_positions
 
    .. automethod:: from_crystfel_geom
+
+   .. automethod:: quad_positions
+
+   .. automethod:: to_h5_file_and_quad_positions
 
    .. automethod:: write_crystfel_geom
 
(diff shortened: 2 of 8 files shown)


tmichela:
I was about to ask if that could be generalized, the I re-read the PR description :)  
At least between DSSC and LPD the differences seem minimal so that should work fine (not sure if agipd would mess things here).
Also, the AGIPD4M will most likely have 2 quadrant. I don't know about geometry for Jungfrau (Are there quadrant-like assembly?), AGIPD500K has 1 quadrant.

AGIPD4M


the code here looks good to me (trusting the tests, I did not run it myself)

takluyver:
Yup, this is also part of why I didn't want to overgeneralise - it's not readily applicable to all detectors. I'm thinking that if you can construct a geometry object from some information (e.g. quadrant positions + XFEL HDF5 geometry), then it should be possible to get that same information back out.

JUNGFRAU doesn't have quadrants, as far as I know. From what I've heard, people seem to use even numbers of modules, so possibly they're assembled in pairs somehow - but at least for now, you need the position of each module to construct a geometry object.

PR #25, opened 2020-09-18
WIP: Geometry class for AGIPD500K2G

tmichela:
AGIPD500K2G_geom

### DIFF
--- a/extra_geom/detectors.py
+++ b/extra_geom/detectors.py
@@ -854,6 +854,261 @@ class AGIPD_1MGeometry(DetectorGeometryBase):
         return super().to_distortion_array(allow_negative_xy)
 
 
+class AGIPD_500KGeometry(DetectorGeometryBase):
+    """Detector layout for AGIPD-500k
+
+    The coordinates used in this class are 3D (x, y, z), and represent metres.
+
+    You won't normally instantiate this class directly:
+    use one of the constructor class methods to create or load a geometry.
+    """
+    detector_type_name = 'AGIPD-500K2G'
+    pixel_size = 2e-4  # 2e-4 metres == 0.2 mm
+    frag_ss_pixels = 64
+    frag_fs_pixels = 64
+    expected_data_shape = (8, 512, 128)
+    n_modules = 8
+    n_tiles_per_module = 16
+
+    @classmethod
+    def from_center(cls, center, asic_gap=2, panel_gap=(16, 30), unit=pixel_size):
+        """Generate an AGIPD-500K2G geometry from center position.
+
+        This produces an idealised geometry, assuming all modules are perfectly
+        flat, aligned and equally spaced within their quadrant.
+
+        The center position is given in pixel units, referring to the first
+        pixel of the first module, corresponding to data channels 0.
+
+        The origin of the coordinates is the bottom-right corner of the
+        detector. Coordinates increase upwards and to the left (looking along
+        the beam).
+
+        To give positions in units other than pixels, pass the *unit* parameter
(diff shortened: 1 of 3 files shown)


extra_geom/detectors.py
```diff
@@ -854,6 +854,276 @@ def to_distortion_array(self, allow_negative_xy=False):
...
+        for p in range(8):
+            panel_corner_y = (center[1] * unit) - ((p // 2) * (module_height + panel_gap_y))
+            panel_corner_x = (center[0] * unit) + ((p % 2) * (module_width + panel_gap_x))
+            # print(f'panel{p} ({1000 * panel_corner_x:.2f}, {1000 * panel_corner_y:.2f})')
+
+            tiles = []
+            modules.append(tiles)
+
+            for a in range(16):
+                # upper ASIC row is rotated by 180 degre
```
takluyver, line 919: It looks like this is going to be fixed upstream (either in AGIPD or the DAQ), so we won't need this in the future. It's not clear yet whether we'll need it as an option for the existing data.

extra_geom/detectors.py
```diff
@@ -854,6 +854,276 @@ def to_distortion_array(self, allow_negative_xy=False):
...
+        max_xy = corners.max(axis=0).astype(int) + 1
+
+        size = max_xy - min_xy
+        centre = -min_xy
+        # Switch xy -> yx
+        return tuple(size[::-1]), centre[::-1]
+
+    @staticmethod
+    def split_tiles(module_data):
+        # 2 rows of 8 ASICs each. This slicing is faster than np.split().
```
takluyver, line 1064: Perhaps this is the point to discuss: do we also need two rows for AGIPD-1M? I initially copied a CrystFEL geometry in defining a module as a single row of 8 panels, 128 x 64 each. I've also seen in the calibration code that it's actually two rows of 64 x 64 ASICs. If there's no gap between the rows, maybe it doesn't matter for the geometry? But we should make sure it's clearly documented in either case.

extra_geom/detectors.py
tmichela: yep, I'll stop working on this for now. I think this won't be necessary if the shape is fixed before the first experiment at HED

extra_geom/detectors.py
tmichela: yep, That's probably fine to consider the 2 asics as 1 tile as long as there is no gap and the data is aligned.

takluyver: Have we got some confirmation that that's how they are?

tmichela: > The dead area of the final system is minimized for each FEM by using a monolithic sensor bump-bonded to 2 × 8 individual ASICs. The large sensor guarantees that there are no blind spots within the sensitive surface (0% dead area); however, since the ASICs cannot physically touch each other, the pixels horizontally in between two ASICs are twice as wide (400 µm × 200 µm), covering the area of two ordinary pixels. Pixels vertically in between two ASICs are of normal size.

https://journals.iucr.org/s/issues/2019/01/00/gb5080/

tmichela: Also (even if that's not a formal proof), there are no gap visible between asics looking at rings

asic

tmichela: Also confirmed here: https://xfel.tind.io/record/395/ (chapter 5)

takluyver: OK, sounds good.

takluyver: & thanks for looking into this :-)

takluyver:
Closing this one, superseded by #26.

PR #26, opened 2020-09-22
FEAT: add support for AGIPD500K2G detector layout

tmichela:
Screenshot from 2020-09-22 17-04-18

Screenshot from 2020-09-22 17-04-31

`module 0` is from recorded data (others are random), you can see the first row of ASICS is rotated by 180 degree


takluyver:
Looks promising. It would be good to have at least a couple of basic tests with this.

Should we use `ax.set_aspect(1)` in inspect() to make the shape clearer? The first plot is stretching the panels vertically quite a bit.

extra_geom/detectors.py
```diff
@@ -854,6 +854,170 @@ def to_distortion_array(self, allow_negative_xy=False):
...
+
+        This produces an idealised geometry, assuming all modules are perfectly
+        flat, aligned and equally spaced within their quadrant.
+
+        The center position is given in pixel units, referring to the first
+        pixel of the first module, corresponding to data channels 0.
+
+        The origin of the coordinates is the bottom-right corner of the
+        detector. Coordinates increase upwards and to the left (looking along
+        the beam).
```
takluyver, line 885: This is a bit confusing. It appears to be saying that you give the location of M0 with respect to the bottom beam-right corner (by M6), but that's not the case.

We must be either giving the location of M0 with respect to the centre (the point where a ray from the sample would hit perpendicular to the detector plane) or vice versa - the position of the centre with respect to some fixed point on the detector hardware. I think the implementation is the former, but calling this `from_center` suggests the latter.

extra_geom/detectors.py
```diff
@@ -854,6 +854,170 @@ def to_distortion_array(self, allow_negative_xy=False):
...
+          If True (the default), x increases to the left, as if you were looking
+          along the beam. False gives a 'looking into the beam' view.
+        """
+        ax = super().inspect(axis_units=axis_units, frontview=frontview)
+        scale = self._get_plot_scale_factor(axis_units)
+
+        # Label modules and tiles
+        for ch, module in enumerate(self.modules):
+            cx, cy, _ = module[4].centre() * scale
+            ax.text(cx, cy, f'Mod{ch}', fontweight='bold',
```
takluyver, line 943: Is it the convention that we start with 0 for this one, not from 1 like the Q1M1 numbers for AGIPD-1M?

extra_geom/detectors.py
```diff
@@ -854,6 +854,170 @@ def to_distortion_array(self, allow_negative_xy=False):
...
+            for t in [0, 7]:
+                cx, cy, _ = module[t].centre() * scale
+                ax.text(cx, cy, 'T{}'.format(t + 1),
+                        verticalalignment='center',
+                        horizontalalignment='center')
+
+        ax.set_title(f'AGIPD-500K2G detector geometry ({self.filename})')
+        return ax
+
+    def _get_dimensions(self):
```
takluyver, line 956: I don't think we actually need this unless we define a method to assemble data by interpolation - we did that for AGIPD-1M, but it's horrendously slow (maybe there's a better way to do it that I didn't find).

extra_geom/detectors.py
tmichela: Good question. I've identified the modules with the channel number for now, I'll ask DET.

extra_geom/detectors.py
tmichela: Thanks, I updated the text, hope it's clearer.

extra_geom/detectors.py
tmichela: I added the `position_modules_interpolate` method

tmichela:
> Looks promising. It would be good to have at least a couple of basic tests with this.
> 
> Should we use `ax.set_aspect(1)` in inspect() to make the shape clearer? The first plot is stretching the panels vertically quite a bit.

Thanks, I fixed the aspect ration and added unit test.

tmichela:
I removed the `position_modules_interpolate`, it wasn't working properly. I'll add it if it's required at some point, but on't think we need it at all...

extra_geom/detectors.py
```diff
@@ -854,6 +856,159 @@ def to_distortion_array(self, allow_negative_xy=False):
...
+        """
+        # Overridden only for docstring
+        return super().to_distortion_array(allow_negative_xy)
+
+    def write_crystfel_geom(self, *args, **kwargs):
+        super().write_crystfel_geom(*args, nquads=1, **kwargs)
+
+    def position_modules_fast(self, *args, **kwargs):
+        data, _ = super().position_modules_fast(*args, **kwargs)
+        return data, (0, 0)  # center is the origin
```
takluyver, line 1009: I don't think this is quite right. The aim of returning the centre here is to say where the origin of the detector layout you've given EXtra-geom falls *in the output array*. It doesn't make sense to hard-code it.

extra_geom/detectors.py
```diff
@@ -854,6 +856,159 @@ def to_distortion_array(self, allow_negative_xy=False):
...
+
+        This produces an idealised geometry, assuming all modules are perfectly
+        flat, aligned and equally spaced within their quadrant.
+
+        The center position is given in pixel units, referring to the first
+        pixel of the first module, corresponding to data channels 0.
+
+        The origin of the coordinates is the bottom-right corner of the
+        first module of the detector (channel 0). Coordinates increase upwards
+        and to the left (looking along the beam).
```
takluyver, line 887: I still think this is a muddle. The whole notion of the 'center' is kind of confusing here. I'll try to explain:

For the most part, what we're interested in is where all the bits of the detector are relative to one another, so we can assemble a complete image. In the big quadrant-based detectors, we use the centre (where the beam goes through a hole) as the origin, and give the quadrant positions relative to that. This detector doesn't have an obvious central feature, and for assembling an image, it doesn't matter where we put the origin.

But for a second level of analysis, the origin does matter. E.g. if you're doing azimuthal integration, you need a point to integrate around. We've referred to the point where the beam passes through the detector, which works for a detector perpendicular to the beam, but more generally, we want the origin to be where a ray from the sample is perpendicular to the detector plane. PyFAI calls this the PONI, or Point Of Normal Incidence.

I'm not exactly sure how to handle this. Maybe we should have a `AGIPD_500K2GGeometry.default()` class method which constructs it with the origin in the centre of the 8 modules (as an arbitrary choice)?  Do we know anything about how it will be positioned for use?

tmichela: I jsut asked. It will be sitting on a bench at the end of the chamber, perpendicular to the beam direction (similar to SPB). Beam will pass probably above or below the detector. Also learned that the 500K agipd is temporary they will install a full AGIPD at some point.

Do you mean to have 2 methods: 1 `default` from the center (or corner, ..) + 1 from the PONI (how we should call that)?

takluyver: I think that might be the best thing. `default()` with no required arguments, so you can assemble an image without worrying about where the detector is, and something like `from_center` or `from_poni`. I can't decide if 'center' is meaningful enough, or 'PONI' is too much jargon. Maybe it's better to use a word people don't know than one which they might guess an incorrect meaning for.

I also wonder if this second method should invert the numbers, i.e. you give the position of the PONI relative to some part of the detector, rather than giving the position of the sensor relative to the PONI.

extra_geom/detectors.py
tmichela: I take now the center of the output array size

extra_geom/detectors.py
tmichela: So, I added a `default` method. with the reference (0, 0) in the bottom-right corner (so pixel indices are always positive), I can change that to the centrer or an other corner if you prefer...
I also changed `from_center` to `from_origin`, is that an OK name for you? (also inverted the numbers: origin position is relative to the detector corner)

extra_geom/detectors.py
takluyver: I still think this is not quite right. I guess 'center' is a misnomer - the idea is to return the location corresponding to the origin in the geometry object, i.e. the PONI. If the caller wants `(height / 2, width / 2)`, they can easily get that from the array shape.

tmichela: Oh, yeah. I agree. I misunderstood what you meant, will change taht

extra_geom/detectors.py
```diff
@@ -854,6 +856,191 @@ def to_distortion_array(self, allow_negative_xy=False):
...
+            (0, 0), asic_gap=asic_gap, panel_gap=panel_gap, unit=unit
+        )
+
+
+    @classmethod
+    def from_origin(cls, origin, asic_gap=2, panel_gap=(16, 30), unit=pixel_size):
+        """Generate an AGIPD-500K2G geometry from origin position.
+
+        This produces an idealised geometry, assuming all modules are perfectly
+        flat, aligned and equally spaced within their quadrant.
```
takluyver, line 900:
```suggestion
        flat, aligned and equally spaced within the detector.
```

extra_geom/detectors.py
```diff
@@ -854,6 +856,191 @@ def to_distortion_array(self, allow_negative_xy=False):
...
+
+        The origin of the coordinates is relative to the bottom-right corner of
+        the detector. Coordinates increase upwards and to the left (looking
+        along the beam).
```
takluyver, line 904:
```suggestion
        The default origin is in the bottom-right corner of the detector.
        Coordinates increase upwards and to the left (looking along the beam).
        The origin parameter shifts the origin by the given (x, y) offset.
        For some kinds of analysis, the origin should be at the point of normal
        incidence, where a ray from the sample is perpendicular to the detector
        plane.
```

extra_geom/detectors.py
```diff
@@ -854,6 +856,191 @@ def to_distortion_array(self, allow_negative_xy=False):
...
+    detector_type_name = 'AGIPD-500K2G'
+    pixel_size = 2e-4  # 2e-4 metres == 0.2 mm
+    frag_ss_pixels = 64
+    frag_fs_pixels = 128
+    expected_data_shape = (8, 512, 128)
+    n_modules = 8
+    n_tiles_per_module = 8
+
+    @classmethod
+    def default(cls, asic_gap=2, panel_gap=(16, 30), unit=pixel_size):
```
takluyver, line 876: Looking at this now, maybe it's clearer to have just `from_origin`, with a default so the `origin` parameter is optional. What do you think?

extra_geom/detectors.py
```diff
@@ -854,6 +856,191 @@ def to_distortion_array(self, allow_negative_xy=False):
...
+        """Snap geometry to a 2D pixel grid
+
+        This returns a new geometry object. The 'snapped' geometry is
+        less accurate, but can assemble data into a 2D array more efficiently,
+        because it doesn't do any interpolation.
+        """
+        if self._snapped_cache is None:
+            snapped = super()._snapped()
+            snapped.centre = np.array(snapped.size_yx) // 2
+        return self._snapped_cache
```
takluyver, line 1041: I think don't override this, so `centre` is calculated the same way as for the other detectors. Maybe we should rename it to something like `poni` or `geom_origin`, but I don't think it fundamentally needs to be different for this detector.

extra_geom/detectors.py
tmichela: Yep, that's good

extra_geom/detectors.py
tmichela: Hum, thinking about it

>  the idea is to return the location corresponding to the origin in the geometry object, i.e. the PONI.

In that case, it should always be `(0, 0)`, right. Or I did not get it... :thinking:

takluyver: Sorry, should have been more specific. It's the origin from the geometry object, in the coordinates of the output array. So, for this class, I'd expect the `from_origin(origin=` parameter - in pixel units from the low-x, low-y corner - to be returned as the centre position in the array as well (but reversed, because array indexing goes (y, x)).

The idea is to tell you where the theoretical origin/PONI is in the assembled image.

tmichela: Aha, thanks. Hope I got that right now :D

extra_geom/detectors.py
```diff
@@ -854,6 +856,171 @@ def to_distortion_array(self, allow_negative_xy=False):
...
+        """
+        # Overridden only for docstring
+        return super().to_distortion_array(allow_negative_xy)
+
+    def write_crystfel_geom(self, *args, **kwargs):
+        super().write_crystfel_geom(*args, nquads=1, **kwargs)
+
+    def position_modules_fast(self, *args, **kwargs):
+        data, _ = super().position_modules_fast(*args, **kwargs)
+        origin = data.shape - self.origin[::-1]
```
takluyver, line 1020: I think this should be unnecessary. What we're trying to achieve is the same logical operation that the other detectors already do to get the origin ('centre') location for this method.

Have you tried just using the base class' `position_modules_fast()` method? Does that not behave as expected? I can't see that it should need overriding.

extra_geom/detectors.py
```diff
@@ -854,6 +856,171 @@ def to_distortion_array(self, allow_negative_xy=False):
...
+    @classmethod
+    def from_origin(cls, origin=(0, 0), asic_gap=2, panel_gap=(16, 30),
+                    unit=pixel_size):
+        """Generate an AGIPD-500K2G geometry from origin position.
+
+        This produces an idealised geometry, assuming all modules are perfectly
+        flat, aligned and equally spaced within the detector.
+
+        The default origin (0, 0) of the coordinates is the bottom-right corner
+        of the detector. If an other coordinate is given as the origin, it is
```
takluyver, line 884:
```suggestion
        of the detector. If another coordinate is given as the origin, it is
```

extra_geom/detectors.py
```diff
@@ -854,6 +856,171 @@ def to_distortion_array(self, allow_negative_xy=False):
...
+        relative to the bottom-right corner. Coordinates increase upwards and
+        to the left (looking along the beam).
+
+        The origin of the coordinates is relative to the bottom-right corner of
+        the detector. Coordinates increase upwards and to the left (looking
+        along the beam).
```
takluyver, line 890:
```suggestion
```

takluyver:
Thanks!

tmichela:
\o/

thanks for the review :)

Issue #21, opened 2020-08-28
Methods to move detector elements


takluyver:
For refining geometry, it's useful to be able to move elements at different levels - whole detector, quadrant, module or tile. EXtra-geom should provide methods to make this easy.

PR #27, opened 2020-09-25
Add geom.offset() method

takluyver:
Move part or all of the detector by a given amount.

Closes gh-21

### DIFF
--- a/docs/geometry.rst
+++ b/docs/geometry.rst
@@ -41,6 +41,8 @@ but this geometry code works with a position for each tile.
 
    .. automethod:: from_crystfel_geom
 
+   .. automethod:: offset
+
    .. automethod:: quad_positions
 
    .. automethod:: write_crystfel_geom
@@ -83,6 +85,8 @@ which this geometry code can position independently.
    .. automethod:: from_h5_file_and_quad_positions
 
    .. automethod:: from_crystfel_geom
+
+   .. automethod:: offset
 
    .. automethod:: quad_positions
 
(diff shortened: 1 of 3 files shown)


takluyver:
Using this in geoAssembler (see https://github.com/European-XFEL/geoAssembler/pull/24) resulted in code like this:

```python
self.exgeom_obj = (self.exgeom_obj_orig
                   .offset(quad_offsets_m[0], modules=q_slices[1])
                   .offset(quad_offsets_m[1], modules=q_slices[2])
                   .offset(quad_offsets_m[2], modules=q_slices[3])
                   .offset(quad_offsets_m[3], modules=q_slices[4])
                   )
```

I'm thinking about whether & how to make it possible to pass all the offsets to a single method, rather than chaining four calls.

takluyver:
I have refined this so it's possible to specify a separate shift per module. So the above example can now be:

```python
self.exgeom_obj = self.exgeom_obj.offset(
    np.repeat(quad_offsets_m, repeats=4)
)
```

tmichela:
The interface is a bit hard to comprehend I think, as the arguments are interdependent and the shift argument can be applied to all modules, certain module or tiles (each of them can be filter as well). Maybe it is worth adding 1 or 2 examples, or making the doc a bit more explicit?

Else, LGTM!

takluyver:
Good point. I've expanded the docstring now - is that clearer?

tmichela:
Nice, thanks,

LGTM

Issue #19, opened 2020-08-28
Assemble image so centre is half-way across array


takluyver:
The image assembly gives an array and a centre coordinate within the array. For some kinds of analysis, it's convenient to have an array where the centre is 1/2 way up and 1/2 way across the array. It's possible to do this by padding the returned array, but it would be useful for EXtra-geom to provide a method to make this more convenient.

PR #31, opened 2020-10-06
Assemble data around the centre of an array

takluyver:
As described in #19, this adds a method to assemble data so that the centre of the detector is centred within the output array, i.e. at `height // 2, width // 2`. I've called this `position_modules_symmetric` for now, but I'd welcome suggestions for a clearer name.

I haven't made a corresponding method to create an empty output array (like `output_array_for_position_fast`), though it's easy enough to add if it's useful. Code that wants to reuse an output array can always get one by assembling an empty stack of data first.

### DIFF
--- a/extra_geom/tests/test_agipd_geometry.py
+++ b/extra_geom/tests/test_agipd_geometry.py
@@ -62,6 +62,10 @@ def test_assemble_symmetric():
     # Smoketest assembling into suitable output array
     geom.position_modules_symmetric(stacked_data, out=img)
 
+    with pytest.raises(ValueError):
+        # Output array not big enough
+        geom.position_modules_symmetric(stacked_data, out=img[:-1, :-1])
+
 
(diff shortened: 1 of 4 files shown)


extra_geom/tests/test_agipd_geometry.py
```diff
@@ -44,6 +44,25 @@ def check_result(img, centre):
         img, centre = geom.position_modules_fast(stacked_data, threadpool=tpool)
         check_result(img, centre)
 
+
+def test_assemble_symmetric():
```
kirienko, line 48: Ideally we should also check that `ValueError` is properly raised too :wink:

kirienko (approved):
LGTM

extra_geom/tests/test_agipd_geometry.py
takluyver: Thanks Yury, I've added a check at the end of this test.

PR #32, opened 2020-10-13
Use XMPL data in examples

takluyver:
This allows more people to run the examples, because they only need access to the generic XMPL data on Maxwell.

The AGIPD run is actually a copy of the run I had used before, so it produces the same images. The LPD data is different, but looks superficially similar. I also got rid of the `clip()` function in the LPD example, in favour of using the `vmin` & `vmax` parameters when plotting.


daviddoji:
Imports are using `karabo_data` instead of `extra-data`.
Apart from that, LGTM

takluyver:
Thanks David, well spotted. I've updated them both to use `extra_data`.

PR #33, opened 2020-10-22
Add function to mask AGIPD ASIC joins

takluyver:
Create a mask for the pixels in AGIPD data which are double the width of normal pixels. This can be used to mask data from those pixels out, or to attempt to correct it.

This is a bit different from the rest of EXtra-geom: it relates to the data the detector captures, not the actual geometry. But it's part of how the physical layout of the detector affects the data it captures, and thus part of what you need to make sense of the data. And it's too trivial to make a new package for, so this seems like the best place for it.

@tmichela from what I read in the [paper you linked to before](https://journals.iucr.org/s/issues/2019/01/00/gb5080/), and what I can see in the (old) data I'm working with, it looks like only the boundaries where ASICs meet are double width, and the ASICs on either end of the module have normal pixels at the outer edge. This is what I've implemented for now - let me know if you think I'm wrong.

Here's one AGIPD module with prominent stripes where double width pixels have collected extra light:

image

And here's the same data but with `nan` where the mask from this function is True:

image

Finally, here it is with the data in those pixels halved:

image

Simply halving the corrected value isn't quite right, as we're now getting dark stripes. Jola said that makes sense given how the calibration pipeline calculates a baseline shift. It might be possible to do better - I think the baseline shift is saved in the corrected files now.


takluyver:
Thoughts on how to fit this into the documentation are welcome.

tmichela:
LGTM

> Thoughts on how to fit this into the documentation are welcome.

Somewhere in the mask example notebook?
https://extra-geom.readthedocs.io/en/latest/masks.html

takluyver:
Close/reopen to try to get RTD pull request builds working.

takluyver:
RTD pull request builds working: https://extra-geom--33.org.readthedocs.build/en/33/masks.html#Mask-AGIPD-wide-pixels

Good idea to put it in that notebook.

tmichela:
:+1: LGTM

PR #35, opened 2020-11-23
Dependabot integration

RobertRosca:
Same as for extra-data: added `.github/dependabot/constraints.txt` which is used to constrain package versions installed during the test, and it is used by dependabot to trigger PRs when a dependency is updated. Also added python 3.9 to the test matrix.


takluyver:
:+1:

PR #36, opened 2020-11-25
Reorganise examples in documentation

takluyver:
The idea is to separate two groups of examples: those about creating/loading geometry for specific detectors, and those about using a geometry object. The methods to create & load geometry vary quite a bit between detectors, but what you can do once you have loaded it is more consistent.

This also has the benefit that the code which uses real data on Maxwell is hopefully confined to the 'assembling images' example.

RTD builds pull requests now, which should be valuable for evaluating this.


takluyver:
ReviewNB is pretty nice for this as well.

tmichela:
LGTM

> ReviewNB is pretty nice for this as well.

That's really useful!

RobertRosca:
Nice to see it's coming in handy! It does a pretty good job and the per-cell comments are also really nice.

PR #41, opened 2020-12-15
FIX: Agipd500k module slow scan read orientation

### LINKED ISSUE #40: Mini-half modules are flipped in geometry
Within the calibration team, we use extra-geom 1.0.0 for rendering figures in reports.
This works well, but the mini-half geometries have their x-axis reversed.

image

This is not that left-right halves are swapped, though.

Currently, we work around the issue by doing a `np.fliplr(data)` prior to creating the figure.  
Here's the same data:

image

I'm happy to make a PR fixing this, with a bit of guidance.

Thanks!

tmichela:
closes #40 

inspect

assemble

@cydanil

### DIFF
--- a/extra_geom/detectors.py
+++ b/extra_geom/detectors.py
@@ -1016,15 +1016,16 @@ class AGIPD_500K2GGeometry(DetectorGeometryBase):
         module_height = cls.frag_fs_pixels * cls.pixel_size
         # In x, 64 px + gap between tiles (asics)
         tile_width = (cls.frag_ss_pixels + asic_gap_px) * cls.pixel_size
-        module_width = 8 * tile_width
+        module_width = 8 * tile_width - asic_gap_px * cls.pixel_size  # 8 tiles + 7 gaps
 
         # coordinates relative to the first pixel of the first module
         # detector's bottom-right corner
         ref = (
-            asic_gap_px * cls.pixel_size - module_width,      # x
+            - module_width,  # x
             - (3 * (cls.frag_fs_pixels + panel_gap[1]) * unit)  # y
         )
-        ref = (- (origin[0] * unit + ref[0]), - (origin[1] * unit + ref[1]))  # origin
+        # origin
+        ref = (- (origin[0] * unit + ref[0]), - (origin[1] * unit + ref[1]))
 
         modules = []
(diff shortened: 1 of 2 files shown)


extra_geom/tests/test_agipd500k2g_geometry.py
```diff
@@ -152,6 +152,6 @@ def test_data_coords_to_positions():
     np.testing.assert_allclose(resz, 0)
     np.testing.assert_allclose(resy, 100 * geom.pixel_size)
 
-    assert (np.diff(resx) > 0).all()   # Monotonically increasing
-    np.testing.assert_allclose(resx[0], -512 * geom.pixel_size)
+    assert (np.diff(resx) < 0).all()   # Monotonically increasing
```
takluyver, line 155:
```suggestion
    assert (np.diff(resx) < 0).all()   # Monotonically decreasing
```

extra_geom/detectors.py
```diff
@@ -1020,7 +1020,10 @@ def from_origin(cls, origin=(0, 0), asic_gap=2, panel_gap=(16, 30),
 
         # coordinates relative to the first pixel of the first module
         # detector's bottom-right corner
-        ref = (0, - (3 * (cls.frag_fs_pixels + panel_gap[1]) * unit))  
+        ref = (
+            asic_gap_px * cls.pixel_size - module_width,      # x
```
takluyver, line 1024: You define module width as 8 tiles plus 8 asic (tile) gaps, so here you're subtracting the extra gap. I'd say module width should be 8 tiles plus *7* asic gaps, so this would just be `- module_width`. Then panel_gap_x would be from the end of one module to the start of its neighbour, without projecting an extra asic gap into it.

takluyver (commented):
One suggestion for clarity/accuracy in the calculations, otherwise LGTM.

extra_geom/tests/test_agipd500k2g_geometry.py
tmichela: :sweat_smile:

extra_geom/detectors.py
tmichela: Thanks, yes that makes sens and I changed it

Issue #14, opened 2020-06-24
Add tests for pnCCD detector class


takluyver:
The code was added in #13 and needed to be used soon, so we temporarily skipped writing  tests. It's a fairly straightforward subclass, but has a small extra feature to reshape the input data from a single image to a stack of 2 modules.

PR #45, opened 2021-01-12
Tests and docs for PNCCDGeometry

### LINKED ISSUE #15: Documentation for pnCCD detector
The code was added in #13, and we didn't document it at the time.

JamesWrigley:
Note: I'm not quite sure which methods should be documented, from looking at the existing docs it doesn't seem to be just the union of methods from `DetectorGeometryBase` and the subclass.

Fixes #15 and #14.

### DIFF
--- a/docs/geometry.rst
+++ b/docs/geometry.rst
@@ -235,3 +235,29 @@ Each module is further subdivided into 8 sensor tiles.
    .. automethod:: inspect
 
    .. automethod:: compare
+
+.. _det-PNCCD:
+
+PNCCD
+-----
+
+PNCCD detectors have two movable modules of 1024×512 pixels, each of which is a
+single tile.
+
+.. autoclass:: PNCCDGeometry
+
+   .. autoattribute:: expected_data_shape
+
+   .. automethod:: from_relative_positions
+
+   .. automethod:: from_absolute_positions
+
+   .. automethod:: plot_data_fast
+
+   .. automethod:: position_modules_fast
+
+   .. automethod:: output_array_for_position_fast
+
+   .. automethod:: inspect
+
+   .. automethod:: compare
--- a/extra_geom/tests/test_pnccd_geometry.py
+++ b/extra_geom/tests/test_pnccd_geometry.py
@@ -1,0 +1,58 @@
+import pytest
+import numpy as np
+from matplotlib.axes import Axes
+
+from .utils import assert_geom_close
+from extra_geom import PNCCDGeometry
+
+def test_positions():
+    # Positional constants
+    gap_pixels = 50
+    offset_pixels = 10
+
+    # Calculate parameters for the relative/absolute methods
+    top_offset = (-offset_pixels * PNCCDGeometry.pixel_size, 0, 0)
+    bottom_offset = (offset_pixels * PNCCDGeometry.pixel_size, 0, 0)
+    absolute_pos = np.array(
+        [(-PNCCDGeometry.frag_fs_pixels // 2 - offset_pixels,
+          PNCCDGeometry.frag_ss_pixels + gap_pixels // 2,
+          0),
+         (-PNCCDGeometry.frag_fs_pixels // 2 + offset_pixels,
+          -gap_pixels // 2,
+          0)]
+    ) * PNCCDGeometry.pixel_size
+
+    # Create the geometries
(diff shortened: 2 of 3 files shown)


docs/geometry.rst
```diff
@@ -235,3 +235,27 @@ Each module is further subdivided into 8 sensor tiles.
...
+
+PNCCD
+-----
+
+PNCCD detectors have two movable modules of 1024×512 pixels, each of which is a
+single tile.
+
+.. autoclass:: PNCCDGeometry
+
+   .. automethod:: from_relative_positions
```
philsmt, line 249:
```suggestion
.. autoclass:: PNCCDGeometry

   .. autoattribute:: expected_data_shape

   .. automethod:: from_relative_positions
```

Since `expected_data_shape` is mentioned in the class' docstring, it might be a good idea to spell it out. It also seems I forgot to add the `:attr:` tag to said property in that docstring.

extra_geom/tests/test_pnccd_geometry.py
```diff
@@ -0,0 +1,60 @@
...
+        assert img.shape == expected_img_shape
+        assert tuple(centre) == (expected_img_shape[0] // 2, expected_img_shape[1] // 2)
+        # Everywhere in the gap should be NaN
+        assert np.isnan(img[img.shape[0] // 2, 0])
+        assert img[50, 50] == 0
+
+    # No re-shaping necessary
+    check_result(PNCCDGeometry.expected_data_shape)
+    # The class should automatically reshape the data
+    check_result((PNCCDGeometry.frag_ss_pixels * PNCCDGeometry.n_modules, PNCCDGeometry.frag_fs_pixels))
```
philsmt, line 53: You mind parametrizing `shape` in this test using `@pytest.mark.parametrize`? It might make the intent a bit more clearer than with nested function.

philsmt:
Thanks a lot for working on this! This should have been my job *hides in a corner*

docs/geometry.rst
JamesWrigley: Gotcha, added in 3aa9613.

extra_geom/tests/test_pnccd_geometry.py
JamesWrigley: Ah yeah, that's much nicer. Fixed in 95b193a.

philsmt:
Spot on, thanks. LGTM from my side, let's give Thomas a moment to look over it.

philsmt (approved):

extra_geom/tests/test_pnccd_geometry.py
```diff
@@ -0,0 +1,59 @@
...
+    # Create the geometries
+    absolute_geom = PNCCDGeometry.from_absolute_positions(top=absolute_pos[0], bottom=absolute_pos[1])
+    relative_geom = PNCCDGeometry.from_relative_positions(gap=gap_pixels * PNCCDGeometry.pixel_size,
+                                                          top_offset=top_offset,
+                                                          bottom_offset=bottom_offset)
+
+    # They should be the same
+    assert all(np.array_equal(abs_geom_mod[0].centre(), rel_geom_mod[0].centre())
+               for abs_geom_mod, rel_geom_mod
+               in zip(absolute_geom.modules, relative_geom.modules))
```
takluyver, line 33: We've got an `assert_geom_close()` utility function which I think could be used here.

("close" there is like numpy and the math module use the word, i.e. essentially equal but with a small tolerance for floating point errors)

takluyver (commented):
Thanks - I found one more small improvement, otherwise this LGTM too.

takluyver:
Oh, and about documenting inherited methods: I've generally tried to document the pieces that we're reasonably confident do something sensible and useful for each detector. But it's possible it's behind the curve in some places.

I also set out to describe each detector class independently, without referring to a common section, based on the idea that a typical user will be working with one detector type, and it's useful to have all the information for that together. But we've gradually moved more functionality to the base class and added more detector types, so there's more duplication now, and we may need to re-evaluate that at some point. On the other hand, autodoc makes it fairly painless to have duplicated method descriptions. :slightly_smiling_face:

extra_geom/tests/test_pnccd_geometry.py
JamesWrigley: Ah cool, changed it to use that function (in the squash).

takluyver (approved):

takluyver:
Thanks James!

PR #49, opened 2021-02-15
Read & write CrystFEL format geometry for JUNGFRAU

takluyver:
This also makes reading CrystFEL format geometry more robust (I hope) by not relying on the `p0a0` panel names (which are arbitrary, as far as CrystFEL is concerned). Instead, we identify which piece of the input data each panel description refers to, by the 'data coordinate' (module number, slow-scan, fast-scan) of its first pixel.

Screenshot from 2021-02-15 17-54-17

### DIFF
--- a/docs/geometry.rst
+++ b/docs/geometry.rst
@@ -222,6 +222,10 @@ Each module is further subdivided into 8 sensor tiles.
 
    .. automethod:: from_module_positions
 
+   .. automethod:: from_crystfel_geom
+
+   .. automethod:: write_crystfel_geom
+
    .. automethod:: get_pixel_positions
 
    .. automethod:: to_distortion_array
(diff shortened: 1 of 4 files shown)


takluyver:
This now also includes writing JUNGFRAU geometry to .geom files, because the easiest way to test is to write+read.

In the process, I fixed the code that generates 'rigid group' definitions in the `.geom` file so it doesn't try to describe quadrants for detectors which don't have quadrants.

I've also made an approximate definition of this JF4M layout. At the moment, this is only in the tests (`jf4m_geometry()` function), but if it works and the same layout is going to be used frequently, we could move it into the package as a more convenient way to create an initial geometry.

takluyver:
Fabio reviewed this, including testing it with a real geometry file and real data.

PR #50, opened 2021-02-17
Make scipy.ndimage a local import

takluyver:
The `position_modules_interpolate` method is rarely used (as far as I know), partly because it's very slow. We needn't load scipy.ndimage every time EXtra-geom is imported.


tmichela:
LGTM

PR #51, opened 2021-02-17
Creation of FAQ

daviddoji:
Initial questions were added.

Contributions are very welcome.


docs/FAQ.rst
```diff
@@ -0,0 +1,17 @@
...
+At EuXFEL, the convention is to show detector images looking at the front of the detector, meaning the beam coming from behind you. You can get the other view with EXtra-geom's plotting methods by passing `frontview=False`.
+
+If you use matplotlib's `imshow(array)` to display an image, it may be upside down. You can correct this by passing `imshow(array, origin='lower')`.
+
+2. What are "slow_scan" and "fast_scan"?
+
+They are directions, or axes, in the HDF5 files. They are not always the same for different detectors. For example:
+
+- In LPD, "slow_scan" corresponds to the y-axis and "fast_scan" to the x-axis.
+- In AGIPD, "slow_scan" is the y-axis, whereas "fast_scan" is the y-axis
```
takluyver, line 15: I think we should try to explain what they mean and why we're not using x & y, rather than making a list of specific detectors. It's not as simple as this suggests - e.g. half the AGIPD-1M modules are 'upside down', so fast scan might be +y or -y depending on which module you look at. There's even an idea of using AGIPD in a 'windmill' arrangement, where fast scan would be x for two quadrants and y for two. I don't know if that will ever happen, but CSPAD (at LCLS) really does have modules in different orientations.

When you index the data in Python or C, fast_scan will be the last index, and slow_scan the one before that: `arr[..., slow_scan, fast_scan]`. (Fortran indexing works the other way round, but we probably don't need to mention that). The fast scan axis is the innermost one, where a row of pixels along that axis is stored together.

We talk about slow-scan and fast-scan rather than x & y, because the axes of the data array don't always represent the same physical directions. Working along the fast-scan axis might be increasing y in one AGIPD module, and decreasing y in another.

We could also point people to the `data_coords_to_positions()` method, which converts (module_no, slow_scan, fast_scan) data coordinates to (x, y, z) positions.

takluyver (commented):
Thanks! Some comments inline. One more minor one: can we make the filename lowercase, for consistency with the others?

docs/faq.rst
```diff
@@ -0,0 +1,18 @@
...
+If you use matplotlib's ``imshow(array)`` to display an image, it may be upside down. You can correct this by passing ``imshow(array, origin='lower')``.
+
+What are "slow_scan" and "fast_scan"?
+-------------------------------------
+
+We talk about "slow_scan" and "fast_scan" rather than just "x" and "y" axes, because the axes of the data array in HDF5 files do not always represent the same physical directions. For example, working along the "fast_scan" axis might be increasing "y" in one AGIPD module, and decreasing "y" in another.
+
+One way of know them it's to use the `geom.inspect()` in EXtra-geom, where you'll see diagrams where the dashed 'first row' lines are always in the "fast_scan" direction.
+
+EXtra-geom provides a convenient method, `data_coords_to_position()<https://extra-geom.readthedocs.io/en/latest/geometry.html?highlight=data_coords_to#extra_geom.AGIPD_1MGeometry.data_coords_to_positions>`_, which converts (module_no, slow_scan, fast_scan) to physical positions.
```
takluyver, line 18:
```suggestion
EXtra-geom provides a method, :meth:`~.AGIPD_1MGeometry.data_coords_to_position`, which converts (module_no, slow_scan, fast_scan) coordinates to physical positions.
```

The `:meth:` syntax makes a cross-link to a method. Sphinx cross links are brilliant - it puts a link to the right place when you build the docs, including doing something useful if you create a PDF or an epub from them.

docs/faq.rst
```diff
@@ -0,0 +1,18 @@
...
+At EuXFEL, the convention is to show detector images looking at the front of the detector, meaning the beam coming from behind you. You can get the other view with EXtra-geom's plotting methods by passing ``frontview=False``.
+
+If you use matplotlib's ``imshow(array)`` to display an image, it may be upside down. You can correct this by passing ``imshow(array, origin='lower')``.
+
+What are "slow_scan" and "fast_scan"?
+-------------------------------------
+
+We talk about "slow_scan" and "fast_scan" rather than just "x" and "y" axes, because the axes of the data array in HDF5 files do not always represent the same physical directions. For example, working along the "fast_scan" axis might be increasing "y" in one AGIPD module, and decreasing "y" in another.
+
+One way of know them it's to use the `geom.inspect()` in EXtra-geom, where you'll see diagrams where the dashed 'first row' lines are always in the "fast_scan" direction.
```
takluyver, line 16:
```suggestion
Fast-scan is the inner dimension of the data, so a row of pixels along the fast-scan dimension are stored together.
One way to see how the dimensions are arranged is to use ``geom.inspect()`` in EXtra-geom. The dashed 'first row' lines in the diagram are always in the fast-scan direction.
```

docs/faq.rst
```diff
@@ -0,0 +1,18 @@
...
+--------------------------------------------
+
+At EuXFEL, the convention is to show detector images looking at the front of the detector, meaning the beam coming from behind you. You can get the other view with EXtra-geom's plotting methods by passing ``frontview=False``.
+
+If you use matplotlib's ``imshow(array)`` to display an image, it may be upside down. You can correct this by passing ``imshow(array, origin='lower')``.
+
+What are "slow_scan" and "fast_scan"?
+-------------------------------------
+
+We talk about "slow_scan" and "fast_scan" rather than just "x" and "y" axes, because the axes of the data array in HDF5 files do not always represent the same physical directions. For example, working along the "fast_scan" axis might be increasing "y" in one AGIPD module, and decreasing "y" in another.
```
takluyver, line 14:
```suggestion
What are "slow-scan" and "fast-scan"?
-------------------------------------

We talk about "slow-scan" and "fast-scan" rather than just "x" and "y" axes, because the axes of the data array in HDF5 files do not always represent the same physical directions. For example, working along the fast-scan axis might be increasing y in one AGIPD module, and decreasing y in another.
```

`fast_scan` in code, but fast-scan as a description in text, IMO.

takluyver (commented):
Thanks David! Don't forget to add it to the table of contents in `index.rst` as well. It can probably go second, just after the geometry reference.

PR #62, opened 2021-04-09
Allow passing xarray DataArray object to position_modules_fast()

### LINKED ISSUE #38: Make "position_all_modules" to accept also xarrays
The method "**position_all_modules**" from the detectors module relies ultimately on the the method "**position_modules**" from the snapped module. As input data, it uses a numpy array with 4 dimensions.

It would be nice to extend the method to xarrays as well, as this is the data structure EXtra-data returns. ATM, I have a very hacky solution that use the **StackView** method from EXtra-data.stacking.

takluyver:
Labelled arrays are expected to have a `module` dimension with labels starting from index 0. Closes #38.

cc @daviddoji

### DIFF
--- a/.github/dependabot/constraints.txt
+++ b/.github/dependabot/constraints.txt
@@ -10,3 +10,4 @@ pyparsing==2.4.7
 python-dateutil==2.8.1
 scipy<1.7.0
 six==1.15.0
+xarray<=0.17.0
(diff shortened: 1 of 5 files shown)


daviddoji:
LGTM. 
Thanks @takluyver.

tmichela (commented):
LGTM

PR #63, opened 2021-04-12
Add example notebook for creating JUNGFRAU geometry

takluyver:


CammilleCC:
LGTM, thanks for adding an example :D

PR #67, opened 2021-04-20
Fix & test assembling from EXtra-data StackView

takluyver:
cc @CammilleCC


CammilleCC:
Tested with `extra-foam`, the tests now pass.
LGTM

PR #69, opened 2021-04-30
Fix calculation of module positions for JUNGFRAU

takluyver:
N tiles have N-1 gaps between them

cc @JunCEEE @dallanto


dallanto:
Thanks Jun for spotting and Thomas for the quick correction!

kirienko (approved):
LGTM, btw :wink:

Issue #64, opened 2021-04-15
a generic class for CrtystFEL geometry


kirienko:
At the moment, `extra-geom` can read CrystFEL `geom` files and produce valid geometries for several types of detectors. However, there are cases when one may need more generic representation than a predefined detector (e.g. for simulation purposes, or for the detector that is not yes supported by `extra-geom`). That's why it might be useful to have a generic class geometry loaded from an arbitrary CrystFEL file. 

What is needed:
- [ ] Detecting how many tiles per module
- [ ] Finding data shapes (`frag_ss_pixels`, `frag_fs_pixels`, `expected_data_shape`)
- [ ] Defining 3 methods which are not implemented in the base class (`split_tiles`, `_tile_slice`, `_module_coords_to_tile`)
- [ ] Possibly, retrieving additional information like:
    * an arbitrary pixel size (`res`[olution] in CrystFEL terms)
    * an arbitrary `data` storage source (other than `/entry_1/instrument_1/detector_1/data`)
    * detector distance (`clen` in CrystFEL terms)
    * `photon_energy`

takluyver:
> an arbitrary data storage source (other than /entry_1/instrument_1/detector_1/data)

This is one place where there's a conceptual difference - `.geom` files describe how to load the data from a file, whereas EXtra-geom doesn't deal with loading data.

JunCEEE:
Hi, how is this coming along? Related to https://github.com/European-XFEL/EXtra-geom/issues/66#issuecomment-822499014, I guess this generic class is exactly what I need.

Can we also implement a method in this class to create an instance directly from a python list or dict of:
- pixel size of each panel
- center positions of pixels on each panel
- clen

I need this badly for SimEx-Lite, please let me know where I can help to implement this class ASAP.

takluyver:
Just to record: the initial version of this was released in version 1.3 ([docs](https://extra-geom.readthedocs.io/en/latest/geometry.html#generic-detector)). Reading CrystFEL geometry files into this generic class has been postponed for now. We'll leave this issue open for that.

I've separately started work on preserving extra info when reading a .geom file and writing it back (branch `crystfel-geom-metadata`) - I need to get back to this and finish it up.

PR #72, opened 2021-05-07
Generic detector

kirienko:
This PR provides a generic interface of an arbitrary unconstrained detector.

There will be two different ways of setting up a detector:
1. ~~From a detector-agnostic `geom`-file in CrystFEL format~~ (partly implemented, will follow in the next PR);
2. From a dictionary of the following form:
```python
pixel_size = 0.001
simple_config = {'pixel_size': pixel_size,
                 'fast_pixels': 128,
                 'slow_pixels': 64,
                 'corner_coordinates': [pixel_size * np.array([5, 1, 0]),
                                        pixel_size * np.array([-133, 5, 0]),
                                        pixel_size * np.array([2, -133, 0]),
                                        pixel_size * np.array([-131, -130, 0])],
                 'n_tiles_per_module': 4,
                 'fs_vec': np.array([0 , 1, 0]),
                 'ss_vec': np.array([1, 0, 0]),
                 'tile_offset': 2 * pixel_size,
                 'tile_vec': [0, -1, 0]
                }
simple = CrystFEL_Geometry(**simple_config)
```

~~Note: this PR is big enough already, even though the part 1) is far from being fully implemented. Maybe I need to move the implementation of `from_crystfel_geom()` into a separate branch. In that case @JunCEEE could start using the basic functionality earlier.~~

I removed `from_crystfel_geom()` from this PR, which reduces the amount of code to review rather dramatically.

When finished, it
solves #64


extra_geom/detectors.py
```diff
@@ -768,6 +768,81 @@ def offset(self, shift, *, modules=np.s_[:], tiles=np.s_[:]):
...
+        `tile_vec`: the direction of tile replication, default=[1, 0, 0]
+
+        """
+        self.pixel_size = pixel_size
+        self.frag_fs_pixels = fast_pixels
+        self.frag_ss_pixels = slow_pixels
+        self.n_modules = len(corner_coordinates)
+        self.n_tiles_per_module = n_tiles_per_module
+        self.expected_data_shape = (self.n_modules,
+                                    self.n_tiles_per_module * self.frag_ss_pixels,
```
takluyver, line 810: You're assuming that the tiles in a module are always arranged in one row along the slow-scan axis. Of the specific detectors we support, this is true for AGIPD, but not for LPD, DSSC or JUNGFRAU. I think we'll want to provide a bit more flexibility here. What about something like this:

```python
GenericDetector(
    ...
    pixels_in_tile = (64, 128),  # slow-scan, fast-scan
    tiles_in_module = (8, 1),    # slow-scan, fast-scan
)

File: extra_geom/detectors.py [hunk 2]
```diff
@@ -768,6 +768,81 @@ def offset(self, shift, *, modules=np.s_[:], tiles=np.s_[:]):
...
+    n_tiles_per_module = None
+
+    def __init__(self, pixel_size: float, fast_pixels: int, slow_pixels: int,
+                 corner_coordinates: [np.ndarray] = [np.zeros(3)],
+                 fs_vec: np.ndarray = np.array([0, 1, 0]),
+                 ss_vec: np.ndarray = np.array([1, 0, 0]),
+                 n_tiles_per_module: int = 1,
+                 tile_offset: float = None,
+                 tile_vec: np.ndarray = None,
+                 ):
```
takluyver, line 792: I'm a bit unsure about this, but if we're aiming for a more *generic* class rather than a *simple* one, we might allow this input with a classmethod like `from_simple_description`, e.g.

```python
geom = GenericDetector.from_simple_description(
    pixel_size=2e-4,
    ...
)
```

This would allow us to have other ways to construct the class if we one day needed more flexibility (e.g. modules arranged in different orientations, like CSPAD).

extra_geom/detectors.py
```diff
@@ -768,6 +768,81 @@ def offset(self, shift, *, modules=np.s_[:], tiles=np.s_[:]):
...
+        DetectorGeometryBase(modules)
+
+    def _tile_slice(cls, tileno):
+        raise NotImplementedError
+
+    def split_tiles(self):
+        raise NotImplementedError
+
+    def _module_coords_to_tile(self):
+        raise NotImplementedError
```
takluyver, line 843: These 3 methods need to be implemented for much of the EXtra-geom functionality to work.

takluyver (commented):
Thanks, this looks like a good start.

extra_geom/detectors.py
```diff
@@ -768,6 +768,81 @@ def offset(self, shift, *, modules=np.s_[:], tiles=np.s_[:]):
...
+            for t in range(n_tiles_per_module):
+                tile = GeometryFragment(corner_coordinates[m] +
+                                        self.tile_vec * t * (tile_offset_value * self.pixel_size + self.tile_offset),
+                                        ss_pixels=self.frag_ss_pixels, fs_pixels=self.frag_fs_pixels,
+                                        ss_vec=ss_vec * pixel_size,
+                                        fs_vec=fs_vec * pixel_size)
+                module += [tile]
+            modules += [module]
+        self.modules = modules
+        DetectorGeometryBase(modules)
```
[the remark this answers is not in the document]
kirienko: I overlooked this, thanks! Fixed.

extra_geom/detectors.py
```diff
@@ -768,6 +768,81 @@ def offset(self, shift, *, modules=np.s_[:], tiles=np.s_[:]):
         ])
 
 
+class CrystFEL_Geometry(DetectorGeometryBase):
```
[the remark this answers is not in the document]
takluyver: Good point! GenericGeometry?

extra_geom/detectors.py
```diff
@@ -768,6 +768,81 @@ def offset(self, shift, *, modules=np.s_[:], tiles=np.s_[:]):
...
+    The coordinates used in this class are 3D (x, y, z), and represent metres.
+    """
+    detector_type_name = 'Generic Detector'
+    pixel_size = None   # should be in meters
+    frag_ss_pixels = None
+    frag_fs_pixels = None
+    expected_data_shape = ()
+    n_quads = 0
+    n_modules = None
+    n_tiles_per_module = None
```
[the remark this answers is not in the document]
kirienko: You're right. Fixed.

extra_geom/detectors.py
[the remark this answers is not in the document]
kirienko: Let it be. Done.

extra_geom/detectors.py
kirienko: Oh, no, maybe that's not true for LPD...

kirienko: I assume that tiles are always arranged in one row along the `tile_vec` vector. I agree that this is a restriction. But with three different vectors we have (`fs_vec`, `ss_vec`, `tile_vec`) one may define rather rich spectrum of detectors, see the documentation I prepared. Allowing tiles to be arranged in 2D will immediately make things much more complex. I assume those types of detectors (LPD-like) may be described via a `geom`-file and loaded with `from_crystfel_geom` function, which I will introduce later.

extra_geom/detectors.py
kirienko: Done

extra_geom/detectors.py
kirienko: @takluyver Can we agree on the current implementation? (See the docs.)

extra_geom/detectors.py
kirienko: I tend to agree, but would we have a _default_ method of construction then?

takluyver: The current classes don't really have a useful *default* constructor (all the examples use things like `AGIPD_1MGeometry.from_crystfel_geometry()` or `LPD_1MGeometry.from_quad_positions()` rather than `LPD_1MGeometry()`). I don't think that's a big problem.

extra_geom/detectors.py
```diff
@@ -768,6 +769,123 @@ def offset(self, shift, *, modules=np.s_[:], tiles=np.s_[:]):
...
+
+        1. the number of modules :attr:`n_modules`
+        2. the number of tiles in a module along the slow-scan direction multiplied by
+           the number of slow-scan pixels per tile :attr:`frag_ss_pixels`
+        3. the number of tiles in a module along the fast-scan direction multiplied by
+           the number of fast-scan pixels per tile :attr:`frag_fs_pixels`
+    """
+    detector_type_name = 'Generic Detector'
+
+    def __init__(self, pixel_size: float, fast_pixels: int, slow_pixels: int,
```
takluyver, line 787: Nitpick: I try to consistently have the slow-scan parameter before the fast-scan parameter - this matches the way you use them when indexing `module_data[slow_scan, fast_scan]`.

extra_geom/tests/test_crystfel_geometry.py
```diff
@@ -0,0 +1,86 @@
+from extra_geom import GenericGeometry
```
takluyver, line 1: Don't foget to rename this test file.

Also, as we're leaving reading `.geom` files for a separate PR, let's leave the files for that test until we need them. :)

extra_geom/detectors.py
```diff
@@ -768,6 +769,123 @@ def offset(self, shift, *, modules=np.s_[:], tiles=np.s_[:]):
...
+                 ):
+        """ Create a generic detector:
+
+        `pixel_size`: the size of a pixel in meters (reversed CrystFEL's `res`)
+        `fast_pixels`, `slow_pixels`: the size of a tile along fast and slow axes
+        `corner_coordinates`: 3D coordinates of the first pixel of each module
+        `n_modules`: the number of modules, default=1
+        `n_tiles_per_module`:  the number of tiles in each module, default=1
+        `tile_gap`: the gap between two tiles, default=pixel_size
+        `tile_vec`: the direction of tile replication, default=[1, 0, 0]
```
takluyver, line 803: I don't find this the easiest thing to follow, but that may be just me. :slightly_smiling_face:

extra_geom/tests/test_crystfel_geometry.py
```diff
@@ -0,0 +1,86 @@
...
+def test__tile_silce(simple):
+    _slice = simple._tile_slice(1)
+    assert isinstance(_slice, tuple)
+    assert isinstance(_slice[0], slice)
+    assert isinstance(_slice[1], slice)
+    assert _slice == (slice(0, 64), slice(128, 256))
+
+
+def test_write_crystfel_geom(simple):
+    simple.write_crystfel_geom('test.geom')
```
takluyver, line 44: Does this actually work already? :slightly_smiling_face:

extra_geom/tests/test_crystfel_geometry.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+def test_inspect(simple):
+    assert isinstance(simple.inspect(), plt.Axes)
+
+
+def test_get_pixel_positions(simple):
+    out = simple.get_pixel_positions()
+    assert isinstance(out, np.ndarray)
+    assert out.shape == simple.expected_data_shape + tuple([3])
```
takluyver, line 59:
```suggestion
    assert out.shape == simple.expected_data_shape + (3,)
```

(this is the usual way to write a tuple of 1 item)

extra_geom/tests/test_crystfel_geometry.py
```diff
@@ -0,0 +1,86 @@
...
+    (30, 130),  <- t=1
+    (60, 200),  <- t=1
+    """
+    slow_scan = np.array([5, 10, 30, 60])
+    fast_scan = np.array([10, 50, 130, 200])
+    tileno, tile_ss, tile_fs = simple._module_coords_to_tile(slow_scan, fast_scan)
+
+    assert all(tileno == [0, 0, 1, 1])
+    assert all(tile_ss == slow_scan)
+    assert all(tile_fs == np.mod(fast_scan, 128))
```
takluyver, line 85: Numpy has some functions for checking arrays, like `np.testing.assert_array_equal` and `np.testing.assert_allclose`. We're already using these elsewhere in the tests. `assert_allclose` allows for a little bit of rounding error for floats.

docs/generic_geometry.ipynb
```diff
@@ -0,0 +1,275 @@
+{
```
takluyver, line 1: <p>I'd suggest leaving autoreload out of the examples - it can be handy for development, but it's a distraction in the docs.</p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/European-XFEL/EXtra-geom/pull/72/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

takluyver, line 1: <p>I think this cell especially would benefit from a style that seems to be fashionable now, where you start a new line after an opening <code>{[(</code> bracket and indent the contents less:</p><p> </p><pre class="ql-syntax" spellcheck="false">simple_config = {
    'pixel_size': pixel_size,
    'corner_coordinates': [
        pixel_size * np.array([5, 1, 0]),
        ...
    ],
    ...
}
</pre><p> </p><p>It would also be good to add a 'now that you've created a geometry object, here are things you can do with it' cell at the end, like the detector-specific example notebooks.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/European-XFEL/EXtra-geom/pull/72/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='12'/>

extra_geom/tests/test_crystfel_geometry.py
kirienko: Yes, it does! We inherit everything from the base class, with some parts redefined (like `self._tile_slice`). From the first glance, the result looks plausible. At least @JunCEEE didn't raised any red flags here. 

I will add more checks when the related reading function is implemented.

extra_geom/detectors.py
kirienko: I naturally sort things alphabetically but I always support any systematic approach. Done.

extra_geom/tests/test_crystfel_geometry.py
kirienko: Done!

extra_geom/tests/test_crystfel_geometry.py
kirienko: I totally forgot about this, thanks! Done.

extra_geom/detectors.py
kirienko: Done, docs updated accordingly.

docs/generic_geometry.ipynb
kirienko: My bad, sorry. Removed.

extra_geom/detectors.py
kirienko: I added a bit more to the documentation. 

I used a 3D vector just because there were some already, and that's why I decided that it'd be natural to follow this vector concept with tiles too. If we would decide to use, say, a string variable instead (like `tile_direction = 'left | right | up | down'`) then it required some cumbersome piece of code to interact with numbers.

extra_geom/detectors.py
```diff
@@ -803,41 +804,41 @@ def __init__(self, pixel_size: float, slow_pixels: int, fast_pixels: int,
...
-        self.n_tiles_per_module = n_tiles_per_module
-        self.tile_gap = tile_gap if tile_gap else pixel_size
-        self.tile_vec = np.array(tile_vec) if tile_vec else ss_vec
+        cls.pixel_size = pixel_size
+        cls.frag_fs_pixels = fast_pixels
+        cls.frag_ss_pixels = slow_pixels
+        cls.n_modules = len(corner_coordinates)
+        cls.n_tiles_per_module = n_tiles_per_module
+        cls.tile_gap = tile_gap if tile_gap else pixel_size
+        cls.tile_vec = np.array(tile_vec) if tile_vec else ss_vec
```
takluyver, line 813: This will change the attributes on the class, rather than on an instance. So it will seem to work at first, but if you do:

```python
g1 = GenericGeometry.from_simple_description(pixel_size=2e-4, ...)
g2 = GenericGeometry.from_simple_description(pixel_size=1e-4, ...)
```

Then both `g1.pixel_size` and `g2.pixel_size` will be 1e-4 afterwards (because it's set `GenericGeometry.pixel_size`).

I think we want to calculate the list of modules, then create an instance and attach the attributes to that:

```python
...
inst = cls(modules)
inst.pixel_size = pixel_size
...
return inst
```

extra_geom/detectors.py
takluyver: Thanks :)

extra_geom/detectors.py
takluyver: I was thinking of just `'slow_scan'` or `'fast_scan'`, and not allowing the situation where the tile vector is in the opposite direction to one of the pixel vectors, because I don't think that's particularly likely to come up. But if you consider this and still prefer specifying a vector, that's OK with me.

kirienko: Yes, I would stick to a vector. (BTW one can even put tiles on a diagonal with (1, 1, 0) :smiley:)

docs/generic_geometry.ipynb
kirienko: Done

extra_geom/detectors.py
kirienko: Damn right! I've never thought about such a case, thanks! Will rework later today.

extra_geom/detectors.py
takluyver: Your scientists were so preoccupied with whether or not they could, they didn't stop to think if they should

:wink:

extra_geom/detectors.py
kirienko: Done. It looks a little clumsy, but seems to work.

extra_geom/detectors.py
kirienko: :see_no_evil:

extra_geom/detectors.py
```diff
@@ -768,6 +769,134 @@ def offset(self, shift, *, modules=np.s_[:], tiles=np.s_[:]):
...
+
+        # Get the tile shift: it is a multiple of either `fast_pixels` or `slow_pixels`
+        tile_offset_value = np.abs(np.inner(fs_vec * fast_pixels + ss_vec * slow_pixels, tile_vec))
+        tile_gap = tile_gap if tile_gap else pixel_size
+
+        n_modules = len(corner_coordinates)
+        for m in range(n_modules):
+            module = []
+            for t in range(n_tiles_per_module):
+                tile = GeometryFragment(corner_coordinates[m] +
```
takluyver, line 822:
```suggestion
        for corner_coord in corner_coordinates:
            module = []
            for t in range(n_tiles_per_module):
                tile = GeometryFragment(corner_coord +
```

extra_geom/detectors.py
```diff
@@ -768,6 +769,134 @@ def offset(self, shift, *, modules=np.s_[:], tiles=np.s_[:]):
...
+        assert np.inner(fs_vec, ss_vec) == 0    # scan vectors are perpendicular
+        assert np.linalg.norm(fs_vec) == np.linalg.norm(ss_vec) == 1    # unit vectors
+
+        # the numbers of tiles per module in the fast- and slow-scan directions respectively:
+        geom.fs_tiles = abs(np.inner(fs_vec, geom.tile_vec)) * n_tiles_per_module or 1
+        geom.ss_tiles = abs(np.inner(ss_vec, geom.tile_vec)) * n_tiles_per_module or 1
+
+        geom.expected_data_shape = (geom.n_modules,
+                                    geom.ss_tiles * geom.frag_ss_pixels,
+                                    geom.fs_tiles * geom.frag_fs_pixels)
```
takluyver, line 849:
```suggestion

        assert np.inner(fs_vec, ss_vec) == 0    # scan vectors are perpendicular
        assert np.linalg.norm(fs_vec) == np.linalg.norm(ss_vec) == 1    # unit vectors

        # the numbers of tiles per module in the fast- and slow-scan directions respectively:
        fs_tiles = abs(np.inner(fs_vec, geom.tile_vec)) * n_tiles_per_module or 1
        ss_tiles = abs(np.inner(ss_vec, geom.tile_vec)) * n_tiles_per_module or 1

        geom.expected_data_shape = (geom.n_modules,
                                    ss_tiles * geom.frag_ss_pixels,
                                    fs_tiles * geom.frag_fs_pixels)
```

I'd stick to the attributes we have for all detectors and in all cases. E.g. `tile_gap` won't be available when you load from a .geom file.

extra_geom/detectors.py
kirienko: Well spotted, thanks!

extra_geom/detectors.py
kirienko: That's fine with me, but mind the usage in the definitions of `fs_tiles` and `ss_tiles`. I did a separate commit for that change.

docs/geometry.rst
```diff
@@ -265,3 +265,12 @@ single tile.
...
    .. automethod:: compare
+
+Generic Detector
+----------------
+
+Generic detector may represent an unconstrained detector of your choice.
+
+.. autoclass:: GenericDetector
+
+   .. autoattribute:: expected_data_shape
```
takluyver, line 276: We should document the `from_simple_description` method, along with the list of methods from the base class which are supported (at least write_crystfel_geom, get_pixel_positions, plot_data_fast, position_modules_fast, output_array_for_position_fast, inspect).

I would guess *compare* also works, but we might want a note that it only makes sense between geometries with the same number & size of tiles.

extra_geom/detectors.py
```diff
@@ -768,6 +769,131 @@ def offset(self, shift, *, modules=np.s_[:], tiles=np.s_[:]):
...
+                                n_tiles_per_module: int = 1,
+                                tile_gap: float = None,
+                                tile_vec: np.ndarray = None,
+                                ):
+        """ Create a generic detector:
+
+        `pixel_size`: the size of a pixel in meters (reversed CrystFEL's `res`)
+        `slow_pixels`, `fast_pixels`: the size of a tile along the slow- and the fast-scan axes
+        `corner_coordinates`: 3D coordinates of the first pixel of each module
+        `ss_vec`, `fs_pixels`: 3D vectors of the slow- and the fast-scan directions in the lab coordinates (the X-axis
```
takluyver, line 801:
```suggestion
        `ss_vec`, `fs_vec`: 3D vectors of the slow- and the fast-scan directions in the lab coordinates (the X-axis
```

This also needs fixing in the notebook :-)

extra_geom/detectors.py
```diff
@@ -768,6 +769,131 @@ def offset(self, shift, *, modules=np.s_[:], tiles=np.s_[:]):
...
+                               points to the left, the Y-axis points up, and the Z-axis goes with the beam).
```
takluyver, line 802:
```suggestion
                               points to the left looking along the beam, the Y-axis points up, and the Z-axis goes with the beam).
```

I try to always clarify which left/right I mean, because different people seem to approach it from opposite sides.

extra_geom/detectors.py
```diff
@@ -768,6 +769,131 @@ def offset(self, shift, *, modules=np.s_[:], tiles=np.s_[:]):
...
+        `n_modules`: the number of modules, default=1
```
takluyver, line 803: This one is not a parameter now (again, same in the notebook)

extra_geom/detectors.py
```diff
@@ -768,6 +769,131 @@ def offset(self, shift, *, modules=np.s_[:], tiles=np.s_[:]):
...
+        `n_tiles_per_module`:  the number of tiles in each module, default=1
+        `tile_gap`: the gap between two tiles, default=pixel_size
```
takluyver, line 805:
```suggestion
        `tile_gap`: the gap between two tiles in metres, default=pixel_size
```

extra_geom/detectors.py
takluyver: This should also make it clear that these are unit vectors, not vectors in metres.

extra_geom/tests/test_generic_geometry.py
```diff
@@ -0,0 +1,85 @@
...
+def test__tile_silce(simple):
+    _slice = simple._tile_slice(1)
+    assert isinstance(_slice, tuple)
+    assert isinstance(_slice[0], slice)
+    assert isinstance(_slice[1], slice)
+    assert _slice == (slice(0, 64), slice(128, 256))
+
+
+def test_write_crystfel_geom(simple):
+    simple.write_crystfel_geom('test.geom')
```
takluyver, line 44:
```suggestion
def test_write_crystfel_geom(simple, tmp_path):
    simple.write_crystfel_geom(tmp_path / 'test.geom')
```

If a test writes a file, it's generally nice to do it in a temporary directory rather than just wherever you're running the tests.

takluyver (commented):
Thanks! `ss_tiles` and `fs_tiles` are still additions for just this class, but I think they could be added more generally if people ever came to expect them, so that's OK.

I'm happy with the implementation. My comments below are mainly about the documentation.

docs/geometry.rst
kirienko: Compare does work indeed! I've checked this already. I can image the use case for *one-module-one-tile detector* as rather usual.

extra_geom/detectors.py
kirienko: Thanks for the suggestion, committed!

extra_geom/detectors.py
kirienko: Thanks for the suggestion, committed!

extra_geom/tests/test_generic_geometry.py
kirienko: Agree, my bad. Thanks!

extra_geom/detectors.py
kirienko: Reworked

docs/geometry.rst
kirienko: I guess I added everything you've mentioned

docs/geometry.rst
```diff
@@ -265,3 +265,26 @@ single tile.
...
+
+   .. automethod:: inspect
+
+   .. automethod:: output_array_for_position_fast
+
+   .. automethod:: plot_data_fast
+
+   .. automethod:: position_modules_fast
+
+   .. automethod:: write_crystfel_geom
```
takluyver, line 290: This is definitely me being picky, but can I ask you to order the methods similarly to the other detectors above?

I've roughly tried to arrange methods like this:

- Classmethods to create a geometry object (`from_*`)
- Methods to convert it into another format, either in memory (`get_pixel_positions` / `to_distortion_array`) or writing files
- Methods for working with image data
- Inspection (`inspect` / `compare`)

takluyver:
I'd suggest that we remove the `.geom` files in the tests which we're not currently using, until we need them. :-)

docs/geometry.rst
kirienko: Sure, no problem. (I've put `from_*` in the beginning, and then just alphabetically.) Done

extra_geom/detectors.py
```diff
@@ -768,6 +769,140 @@ def offset(self, shift, *, modules=np.s_[:], tiles=np.s_[:]):
...
+        __________
+
+        pixel_size: float
+            the size of a pixel in meters (reversed CrystFEL's `res`)
+        slow_pixels, fast_pixels: int
+            the size of a tile along the slow- and the fast-scan axes
+        corner_coordinates: ndarray
+            3D coordinates of the first pixel of each module
+        ss_vec, fs_vec: ndarray
+            3D vectors of the slow- and the fast-scan directions in the lab coordinates (the X-axis
```
takluyver, line 808: I still think this should be clearer that these are in units of pixel size, i.e. magnitude 1. 'Lab coordinates' implies they're in regular units like metres (at least to my mind).

kirienko: I tried to fix this, please see. I also removed unused `geom`-files from test directory.

takluyver:
Thanks Yury, this LGTM. You can merge it yourself if there's nothing more you want to do in this PR. Then we can either make a new release with this addition, or look at reading CrystFEL geometry first.

Issue #10, opened 2020-02-24
What do the origin & coordinate axes represent?


takluyver:
The origin for physical coordinates in EXtra-geom is meant to be a meaningful centre point, e.g. for azimuthal integration. But we've been a bit vague so far on what exactly that is: detector centre, the point where the beam intersects the detector plane, or the point where a ray from the interaction point would be normal to the detector plane (PyFAI's 'PONI' or Point Of Normal Incidence).

Also, should the x, y, z directions always be in the lab coordinate system (where +z is along the beam), or based on the detector, so that x & y are in the detector plane, and z only measures deviations from that plane.

When the detector is directly downstream and perpendicular to the beam, these options are equivalent, so it's not important to distinguish between them. But if a detector can be used at an angle, it will quickly become important.

I think that the most practical option is to use the PONI and coordinates based on the detector plane, so that it's always easy to see the detector layout by ignoring the z dimension. But we might want ways to translate between detector-local geometry and lab coordinates.

Prompted by discussion with @dallanto on #6.

kirienko:
That's a really good question. 

On the one hand, the `PONI` coordinate system is easier to maintain, since we can forget about the third Z-axis. 

But on the other hand, if we want `extra-geom` to be compliant with CrystFEL software, we have to either use the [laboratory coordinate system](https://www.desy.de/~twhite/crystfel/manual-crystfel_geometry.html) from the beginning, or to call some special translation function whenever we call `from_crystfel_geom`.

dallanto:
Hi both, excuse my ignorance (which did not harm for my purposes so far) but I've always - weakly - assumed the z-Direction would be in the ``pNaN/coffset`` parameter for each panel, and I could ignore it in practice.  
If so, wouldn't that be a deviation from the plane rather than a ``z`` in lab coords? For me, ``z`` in lab coords would mean something like the sample - pixel distance ...  
But if not (if coffset is something entirely different), where would the ``z`` appear in the panel description at all?

dallanto:
I forgot to add, I'm talking about the CrystFEL geom file here

takluyver:
As things stand, the z coordinate of a panel corner does get translated to coffset when we write a .geom file:

https://github.com/European-XFEL/EXtra-geom/blob/ac76cc5d71f1f7bb7510f08fdf9f3092dbda3829/extra_geom/crystfel_fmt.py#L74-L76

[...]

PR #73, opened 2021-05-21
Add support for ePix100 detector layout

egorsobolev:
Hi there,
this PR adds the somewhat simple geometry of the [ePix100 detector](https://in.xfel.eu/readthedocs/docs/epix-documentation/en/latest/) used at the MID instrument.

It has only a single module. The module built with 4 ASICs with wide pixels on the inner edges.


extra_geom/detectors.py
```diff
@@ -2295,3 +2295,203 @@ def position_modules_fast(self, data, *args, **kwargs):
...
+    # oooo-  -oooo
+    #
+    # o : normal pixels 50x50 um
+    # - : long horizontal pixels 50x175 um
+    # | : long vertical pixels 175x50 um
+    # # : big square pixels 175x175 um (4 big pixels in the center)
+    #
+    # See also
+    # --------
+    # 1. [SLAC Confluence](https://confluence.slac.stanford.edu/display/PSDM/EPIX10KA)
```
takluyver, line 2356: Does it matter that that page appears to be about ePix10k, rather than ePix100? Should this be flagged?

Other pages on that wiki (e.g. [this one](https://confluence.slac.stanford.edu/display/PSDM/EPIX10KA2M+and+EPIX10KAQUAD)) look like epix detectors can be used in multi-module configurations. There's no need to deal with that now if the single module machinery is useful to us, but we might need to deal with it one day.

extra_geom/detectors.py
```diff
@@ -2295,3 +2295,203 @@ def position_modules_fast(self, data, *args, **kwargs):
...
+    # 2. [A Dragone et al 2014 J. Phys.: Conf. Ser. 493 012012]
+    #     (https://doi.org/10.1088/1742-6596/493/1/012012)
+    detector_type_name = 'EPIX100'
```
takluyver, line 2359:
```suggestion
    detector_type_name = 'ePix100'
```

This doesn't need to be all caps - we just have lots of detectors with all caps names.

docs/geometry.rst
```diff
@@ -266,6 +266,32 @@ single tile.
...
+
+   .. automethod:: plot_data_fast
+
+   .. automethod:: position_modules_fast
+
+   .. automethod:: output_array_for_position_fast
+
+   .. automethod:: inspect
+
+   .. automethod:: compare
```
takluyver, line 293: If they're meant to be public, the static methods for dealing with the larger pixels in the 'seams' should be listed as well.

extra_geom/tests/test_epix100_geometry.py
```diff
@@ -0,0 +1,84 @@
...
+def test__tile_silce(epix):
+    _slice = epix._tile_slice(0)
+    assert isinstance(_slice, tuple)
+    assert isinstance(_slice[0], slice)
+    assert isinstance(_slice[1], slice)
+    assert _slice == (slice(352, 704), slice(384, 768))
+
+
+def test_write_crystfel_geom(epix, tmp_path):
+    epix.write_crystfel_geom(tmp_path / 'test.geom')
```
takluyver, line 39: As you've included `from_crystfel_geom` in the docs, can we test loading the file back and checking that we get a similar geometry? `test_jungfrau_geometry` has the neatest existing example of this.

takluyver (commented):
Thanks, this looks to be basically in good shape; just a few comments.

docs/epix100_geometry.ipynb
```diff
@@ -0,0 +1,157 @@
+{
```
takluyver, line 1: <p>I'd normally expect the first pixel to be on one of the outside corners, not in the centre. Unless the data is stored in a really weird way, I think the ASICs are in the wrong orientation.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/European-XFEL/EXtra-geom/pull/73/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='3'/>

extra_geom/detectors.py
egorsobolev: Fine! I thought it was such an agreement. Fixed

docs/geometry.rst
egorsobolev: Fixed

extra_geom/tests/test_epix100_geometry.py
egorsobolev: make sence. done.

docs/epix100_geometry.ipynb
egorsobolev: Good question. I slice them in the order they labeled in docs. But I don't flip them as it did the controller when assembling data in the single array.

I just didn't think that anyone really needs ASICs flipped to match the electronic layout. Do you think it is necessary?

takluyver: How is the data stored in the files? The 'first pixel' should correspond to the data you access with `frame[0, 0]` (and the first row is from there along `frame[0, 1]`, `frame[0, 2]` etc.).

docs/epix100_geometry.ipynb
```diff
@@ -0,0 +1,157 @@
...
+    "geom = Epix100Geometry.from_origin([0, 0])\n",
+    "geom.inspect()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "cfa9432d",
+   "metadata": {},
+   "source": [
+    "We can also load JUNGFRAU geometry from a [CrystFEL format geometry file](https://www.desy.de/~twhite/crystfel/manual-crystfel_geometry.html). These are text based files, often saved with a .geom extension."
```
kirienko, line 74: Epix, not Jungrau, right? :smiley:

docs/epix100_geometry.ipynb
```diff
@@ -0,0 +1,157 @@
...
+    "geom2 = Epix100Geometry.from_crystfel_geom('epix100_example.geom')\n",
+    "geom2.inspect()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "fb70ece4",
+   "metadata": {},
+   "source": [
+    "The code above creates a geometry object for JUNGFRAU. See these examples for how to use a geometry object:\n",
```
kirienko, line 128:
```suggestion
    "The code above creates a geometry object for Epix100. See these examples for how to use a geometry object:\n",
```

kirienko (changes requested):
image

It looks like the directions of axes is opposite to what user expects for the rest of `extra-geom` detectors. Normally, we assume that the X-axis points to the left looking along the beam, the Y-axis points up, and the Z-axis goes with the beam. I.e. (x, y, z) is a right-handed triple.

takluyver:
@kirienko I think the origin shift you show is as I would expect, though this can certainly be somewhat confusing. We have the same `from_origin` method for AGIPD 500k. You're shifting the *origin* itself 100 px up (+y) and beam-left (+x) from its default location, which has the same effect as shifting the detector in the opposite direction.

kirienko:
> @kirienko I think the origin shift you show is as I would expect, though this can certainly be somewhat confusing. We have the same `from_origin` method for AGIPD 500k. You're shifting the _origin_ itself 100 px up (+y) and beam-left (+x) from its default location, which has the same effect as shifting the detector in the opposite direction.

Aha, I see the logic now! But it's still somewhat confusing for me, you're got it right. Because this implies _some_ "detector coordinate system" already established for the Epix, even _before_ we define our lab coordinate system (or whatever, see #10). And then we want to move the origin of our `extra-geom`'s coordinate system relatively to that "detector" coordinate system that we had _implicitly_ defined before.

kirienko:
I essentially mean that won't it be easier to to understand the geometry, if we say: «Here is our coordinate system, and in these coordinates, the center of the detector is over there»?

But once again we face the #10, because we have to define our coordinates first.

takluyver:
IIRC, the logic for AGIPD 500k was that you could tell EXtra-geom where the beam was relative to the detetor. Maybe, now that we have the `.offset()` method (which shifts the way you expect, moving the panels relative to the origin), we should leave 'origin' out of the constructor, so you would do `Epix100Geometry.default().offset((-100 * px_size, -100 * px_size))` instead of `Epix100Geometry.from_origin((100, 100))`. :shrug: 

Unfortunately, every way I can find of describing positions is confusing for at least some use cases (are you measuring a->b or vice-versa, pixels or metres, corner with the first data pixel or bottom beam-right corner, starting from the centre or from a corner...)

kirienko:
Okay, it's more complicated than I thought. Probably we can leave it as is. Maybe, if this feature is used in the wild, we could add the case of non-zero coordinates in `.from_origin([x, y])` to the gallery of examples. I don't think we can solve this issue constructively without a particular definition of the coordinate system in #10.

docs/epix100_geometry.ipynb
egorsobolev: There are certain conventionality in how to slices tiles. You can do this to match the labeling accepted in the documentation, or align the ASIC electrical circuits, or as you suggest be led by packing tiles in data.

I have fixed the code to match the last one. And added support of ePix10k

docs/epix100_geometry.ipynb
egorsobolev: thanks, fixed

docs/epix100_geometry.ipynb
egorsobolev: fixed

extra_geom/detectors.py
```diff
@@ -2295,3 +2295,283 @@ def position_modules_fast(self, data, *args, **kwargs):
...
+    #     two versions
+    # See also
+    # --------
+    # 1. [EuXFEL ePix documentation]
+    #     (https://in.xfel.eu/readthedocs/docs/epix-documentation/en/latest/index.html)
+    # 2. [A Dragone et al 2014 J. Phys.: Conf. Ser. 493 012012]
+    #     (https://doi.org/10.1088/1742-6596/493/1/012012)
+    # 3. [SLAC Confluence](https://confluence.slac.stanford.edu/display/PSDM/EPIX10KA)
+    #     Describes ePix10KA instead of ePix100, many details are common for these
+    #     two versions
```
tmichela, line 2383:
```suggestion
    # See also
    # --------
    # 1. [EuXFEL ePix documentation]
    #     (https://in.xfel.eu/readthedocs/docs/epix-documentation/en/latest/index.html)
    # 2. [A Dragone et al 2014 J. Phys.: Conf. Ser. 493 012012]
    #     (https://doi.org/10.1088/1742-6596/493/1/012012)
    # 3. [SLAC Confluence](https://confluence.slac.stanford.edu/display/PSDM/EPIX10KA)
    #     Describes ePix10KA instead of ePix100, many details are common for these
    #     two versions
```

extra_geom/detectors.py
```diff
@@ -2295,3 +2295,283 @@ def position_modules_fast(self, data, *args, **kwargs):
...
+    frag_ss_pixels = 352  # rows
+    frag_fs_pixels = 384  # columns
+    expected_data_shape = (
+        EpixGeometryBase.n_modules,
+        2*frag_ss_pixels,
+        2*frag_fs_pixels)
+
+    @staticmethod
+    def split_tiles(module_data):
+        return Epix100Geometry._split_tiles(module_data)
```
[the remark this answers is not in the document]
egorsobolev: You are right. For some reason, I thought this wouldn't work. Fixed.

PR #74, opened 2021-05-25
inspect function for GenericGeometry

kirienko:
Implementation of the `inspect` function for `GenericGeometry`. I used the same style as for `AGIPD`, the only difference is that we don't have **quadrants** here.

I also added the possibility to change the aspect ratio of the figure, exactly in the same manner as in `matplotlib.axes.Axes`.

### DIFF
--- a/extra_geom/detectors.py
+++ b/extra_geom/detectors.py
@@ -885,12 +885,17 @@ class GenericGeometry(DetectorGeometryBase):
 
         # Label modules and tiles
         for ch, module in enumerate(self.modules):
+            label = f"M{ch}" if len(self.modules) > 1 else ""
             if self.n_tiles_per_module > 1:
                 for t in [0, self.n_tiles_per_module - 1]:
                     cx, cy, _ = module[t].centre() * scale
-                    ax.text(cx, cy, 'M{}T{}'.format(ch + 1, t + 1),
+                    ax.text(cx, cy, label + f"T{t}",
                             verticalalignment='center',
                             horizontalalignment='center')
+            elif label:
+                # one tile per each of multiple modules: only modules are labelled
+                cx, cy, _ = module[0].centre() * scale
+                ax.text(cx, cy, label, verticalalignment='center', horizontalalignment='center')
         ax.set_aspect(aspect)
         return ax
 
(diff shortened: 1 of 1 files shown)


extra_geom/detectors.py
```diff
@@ -857,6 +857,43 @@ def from_simple_description(cls, pixel_size: float, slow_pixels: int, fast_pixel
...
+        """
+        ax = super().inspect(axis_units=axis_units, frontview=frontview)
+
+        ax.set_title(f"{self.detector_type_name} geometry ({self.filename})")
+
+        scale = self._get_plot_scale_factor(axis_units)
+
+        # Label modules and tiles
+        for ch, module in enumerate(self.modules):
+            if self.n_tiles_per_module > 1:
```
takluyver, line 888: Do you want to add labels with just the module numbers for detectors with 1 tile per module?

extra_geom/detectors.py
```diff
@@ -857,6 +857,43 @@ def from_simple_description(cls, pixel_size: float, slow_pixels: int, fast_pixel
...
+                for t in [0, self.n_tiles_per_module - 1]:
+                    cx, cy, _ = module[t].centre() * scale
+                    ax.text(cx, cy, 'M{}T{}'.format(ch + 1, t + 1),
```
takluyver, line 891: Up to you whether it counts from 1 or 0. For the Q1M1 numbers, starting at 1 seems to be the norm, but when we're just counting modules, there are different conventions in different places.

extra_geom/detectors.py
kirienko: You're right, I overlooked this case! Thanks!
I changed to the following: 
* one module, one tile: no labels at all; 
* one module, `n+1` tiles: `T0 ... Tn`
* `n+1` modules, one tile per module: `M0`, `M1`, ... `Mn`
* many modules, many tiles: `M0T0 ... M0Tn`, ... `MkT0 ... MkTn`

extra_geom/detectors.py
takluyver: That's fine - I think either way it's going to be what some people expect and not what others expect. I think I'd go with starting from zero. We can always add an option later if people complain enough.

kirienko:
Thanks for the review!

PR #75, opened 2021-05-28
Split base class from detector subclasses

takluyver:
@kirienko pointed out that the `detectors` module was getting very long. This cuts it by about 500 lines by separating out the base class.

It's still pretty long, though quite a bit of that is verbose docstrings. We could put each detector class in its own file - but I know that loading a lot of files tends to be slow on GPFS. I'm not sure where the best trade-off is.


kirienko (approved):
LGTM

PR #76, opened 2021-05-31
Housekeeping

kirienko:
I did a bit of housekeeping on top of @takluyver's #75 
Nothing is really important, feel free to reject.

Essentially, there were only three actions performed:
* `isort`ing imports
* ~~`Epix` classes are put in between `DSSC` and `Jungfrau`~~
* `__all__` in the `detectors.py` looked like a leftover (it's now in the `__init__.py`), I removed it

Also a couple of lines were added to make in more `pep8`-compliant.


extra_geom/__init__.py
```diff
@@ -35,10 +35,10 @@
...
 __version__ = "1.3.0"
 
-from .detectors import (
-    AGIPD_1MGeometry, AGIPD_500K2GGeometry, GenericGeometry, DSSC_1MGeometry, JUNGFRAUGeometry,
-    LPD_1MGeometry, PNCCDGeometry, Epix100Geometry, Epix10KGeometry, agipd_asic_seams,
-)
+from .detectors import (AGIPD_1MGeometry, AGIPD_500K2GGeometry,
+                        DSSC_1MGeometry, Epix10KGeometry, Epix100Geometry,
+                        GenericGeometry, JUNGFRAUGeometry, LPD_1MGeometry,
+                        PNCCDGeometry, agipd_asic_seams)
```
takluyver, line 41: I prefer the previous style (with names on 2 lines isntead of 4) on this one.

takluyver (commented):
Thanks! I'm fine with ordering the imports and removing the outdated `__all__`.

I'm not convinced about rearranging the classes, though - they're not in alphabetical order in any case. The ordering is somewhere inbetween 'detectors with similar technical details together' (AGIPD/LPD/DSSC have a common data format) and 'the order in which we added them. I'd be inclined to just leave them as they are rather than try to impose an order.

extra_geom/__init__.py
kirienko: Yep, no problem, done

kirienko:
@takluyver I returned `Epix` classes back to their initial position

takluyver:
Thanks Yury! I'm just going to close & reopen to see if Github will update the diff now that I've merged #75.

kirienko:
Yes, this looks strange. I could try to pull already merged #75 from the master branch now, and see if it helps.

takluyver:
No, that's OK. I've checked the diff locally, and it's alright, it's just the view on Github.

PR #77, opened 2021-05-31
GenericGeometry.from_crystfel_geom is not implemented

kirienko:
This is just a stub. We want user to know that the `GenericGeometry.from_crystfel_geom` method is not implemented yet.

Tests are aware of this fact.


takluyver:
Thanks Yury, LGTM.

PR #78, opened 2021-06-10
Use np.zeros() to make output array for integer data

takluyver:
We've found elsewhere that filling integer data with NaN doesn't always convert it to 0. `np.zeros()` is also much faster than `np.full(..., 0)` - largely because of virtual memory tricks which defer some of the cost until you write to it, but it's probably still a useful optimisation for big arrays. Here's a blog post with more background about these memory tricks: https://vorpus.org/blog/why-does-calloc-exist/

cc @daviddoji

### DIFF
--- a/extra_geom/snapped.py
+++ b/extra_geom/snapped.py
@@ -84,7 +84,12 @@ class SnappedGeometry:
         """Make an output array for self.position_modules()
         """
         shape = extra_shape + self.size_yx
-        return np.full(shape, np.nan, dtype=dtype)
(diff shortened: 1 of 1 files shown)


takluyver:
For the record, this came up when assembling LPD raw data (uint16). Testing with 3500 images:

* Before this PR: roughly 6.5 seconds serial, 3.3 seconds with 10 threads
* With this PR: roughly 6.0 seconds serial, 1.2 seconds with 10 threads

daviddoji:
Thanks @takluyver. LGTM!

PR #87, opened 2021-07-15
Store some metadata from .geom files and write it back if not superseded.

takluyver:
Specifically, this stores the `data` & `mask` HDF5 paths, `adu_per_eV` and `clen`. I'd like to store `photon_energy` as well, but it appears cfelpyutils doesn't give us that at present. I'll try to get that fixed.

There's a small wrinkle: all of these properties can be specified per panel, although it seems to be common to specify them once for the entire detector. cfelpyutils gives us the numbers per panel, even if they were specified for the whole detector. Possible options to deal with this:

- Assume the values on the first panel apply to the whole detector (what this does now for simplicity)
- Only store the values if they're the same for all panels
- Store & write out separate values for every single panel (at the cost of some further complexity)

cc @kirienko @JunCEEE @dallanto @turkot


takluyver:
It appears the `photon_energy` thing is already fixed in cfelpyutils development version, but not yet released.

JunCEEE:
As long as there is no urgent user request for this, I think it's okay to keep the current simple way. To have the values separately for each panel will only be useful if one is dealing with the raw data of AGIPD.  But I don't think people will deal with raw data with extra-geom.

extra_geom/base.py
```diff
@@ -311,6 +326,20 @@ def write_crystfel_geom(self, filename, *,
         """
         if nquads is None:
             nquads = self.n_quads
+
+        # If this geometry came from a .geom file, restore metadata from there
+        cfelmeta = self.metadata.get('crystfel', {})
+        if data_path is None:
+            data_path = cfelmeta.get('data') or '/entry_1/instrument_1/detector_1/data'
```
tmichela, line 333:
```suggestion
            data_path = cfelmeta.get('data', default='/entry_1/instrument_1/detector_1/data')
```

tmichela (commented):
LGTM

extra_geom/base.py
takluyver: The advantage of using `or` is that it will use the default if the dict contains something like `None` - which it might, from the way we construct it (`{k: d1.get(k) for k in cfel_md_keys}`).

PR #89, opened 2021-08-03
Add method to make DSSC-1M geometry from only quadrant positions

takluyver:
Similar to capabilities we already have with other quadrant-based detectors. It's less accurate than getting geometry from a file, but sometimes it's convenient not to need a file.


extra_geom/detectors.py
```diff
@@ -1014,6 +1014,90 @@ class DSSC_1MGeometry(DetectorGeometryBase):
...
+        unit: float, optional
+          The conversion factor to put the coordinates into metres.
+          The default 1e-3 means the numbers are in millimetres.
+        asic_gap: float, optional
+          The gap between adjacent tiles/ASICs. The default is 2 mm.
+        panel_gap: float, optional
+          The gap between adjacent modules/panels. The default is 4 mm.
+        """
+        assert len(quad_pos) == 4
+        asic_gap_m = 2e-3 if (asic_gap is None) else asic_gap * unit
```
kirienko, line 1045: IMO, that's not quite obvious: in the function definition we say `asic_gap=None` and `panel_gap=4` but we _implicitly_ define the default `asic_gap` is not `None` but two mm. While with the `panel_gap` value we define the default value explicitly (well, we do another trick which leads to that) as four mm. Maybe there is a reason behind this way of definition, but otherwise I'd think about the way of defining both `asic_gap` and `panel_gap` in a more, so to say, symmetric way.

kirienko (approved):
I didn't check the rendered jupyter notebook, but the rest looks fine :smiley: (I still interested, yes). Also see my comment regarding the default values of gaps.

extra_geom/detectors.py
takluyver: Good catch, thanks. They should both be None, so that if you pass `unit=`, the default gaps are still 2 mm & 4 mm, not 2 & 4 of whatever unit you specify.

extra_geom/detectors.py
```diff
@@ -1014,6 +1014,90 @@ class DSSC_1MGeometry(DetectorGeometryBase):
         [0.5, 1, 1, 0.5, 0, 0]
     ])
 
+    @classmethod
+    def from_quad_positions(cls, quad_pos, *, unit=1e-3, asic_gap=None,
+                            panel_gap=None):
```
tmichela, line 1019: I realised now that we are a bit inconsistant in the default values of asic and panel gap, sometimes it's `None` and sometime it's the default value.

tmichela (commented):
LGTM

extra_geom/detectors.py
takluyver: We are, yup. My take is that it should be None wherever `unit` affects how the gaps are interpreted (which I think is everywhere you can pass them), so you can specify `unit` without also having to specify the gaps. We should go through and make that consistent.

PR #92, opened 2021-08-11
Store & use quadrant positions in EuXFEL HDF5 format geometry files

### LINKED ISSUE #91: Support quadrant positions in h5 files
It would be nice to write the quadrant positions to the hdf5 file in `to_h5_file_and_quad_positions()` and at the same time make the `positions ` parameter of `from_h5_file_and_quad_positions()` optional if the positions exist in the file (while still allowing overwriting the quad positions in the file if they are passed explicitly).

takluyver:
`to_h5_file_and_quad_positions` will also write quadrant positions into the file, in a similar format to the existing module & tile positions. There is a new class method `from_h5_file` to load such a geometry file without passing in quadrant positions separately.

Quadrant positions are relative to the centre of the detector (often the point where the beam passes through, but see #10). This is what EXtra-geom uses internally, inspired by CrystFEL geometry, but might be surprising for DET.

For now, this only applies to LPD & DSSC, as these are the only detectors where we've implemented reading & writing HDF5 files.

Closes #91.

### DIFF
--- a/docs/geometry.rst
+++ b/docs/geometry.rst
@@ -132,6 +132,8 @@ which this geometry code can position independently.
 
    .. automethod:: from_h5_file_and_quad_positions
 
+   .. automethod:: from_h5_file
+
    .. automethod:: from_crystfel_geom
 
    .. automethod:: offset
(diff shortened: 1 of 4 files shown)


tmichela:
LGTM

FilipeMaia:
👍

Issue #96, opened 2021-08-24
ePix100 : wrong expected data shape


tmichela:
[documentation](https://iopscience.iop.org/article/10.1088/1742-6596/493/1/012012) says ePix100 asic's Array size is 352x384, but the data we store has 2 extra rows.

takluyver:
Is that related to this note in the docstring?

> There are 4 more rows in raw data. These are calibration pixels. They have the same electronics as normal pixels but aren't wired to the sensor. They are two first and two last rows in the raw data array. This class assumes that calibration rows are cut.

cc @egorsobolev

egorsobolev:
Yes, it is. Our raw data contain these calibration rows.

tmichela:
closed by #97

PR #97, opened 2021-08-24
ePix100: allow setting different asic gaps on x and y axis

### LINKED ISSUE #95: ePix100 : Allow different asic gap size
There are two geometries used with ePix100

1) a single monolithic sensor with a 2x2 array of four ASICs bonded to it.
These would have no dead gaps but would have large pixels in the central cross
-> current implementation if gaps

2) A pair of sensors with each sensor being bonded to two ASICs.
These would have a dead gap equal to twice the guard ring width (~450-500um)
plus a mechanical gap of about 200-300 microns. This would resutl in a total
dead gap of about 1.25 millimeters.

tmichela:
closes #95 #96 

example data with asymetric gaps (see #95 ):

epix100_plot_gaps

@takluyver @egorsobolev

### DIFF
--- a/extra_geom/detectors.py
+++ b/extra_geom/detectors.py
@@ -1,7 +1,7 @@
 """Detector geometry handling."""
 import warnings
 from itertools import product
-from typing import List, Tuple
+from typing import Collection, List, Tuple
 
 import h5py
 import numpy as np
@@ -1682,19 +1682,24 @@ class EpixGeometryBase(DetectorGeometryBase):
         if unit is None:
             unit = cls.pixel_size
         if asic_gap is None:
-            asic_gap = cls.asic_gap
+            asic_gap_h = asic_gap_v = cls.asic_gap
+        elif isinstance(asic_gap, Collection):
+            asic_gap_h = asic_gap[0]
+            asic_gap_v = asic_gap[1]
+        else:
+            asic_gap_h = asic_gap_v = asic_gap
 
         x0, y0 = origin[0] * unit, origin[1] * unit
         tiles = []
-        gap = asic_gap * unit
-        row_sz = cls.frag_ss_pixels * cls.pixel_size + gap
-        col_sz = cls.frag_fs_pixels * cls.pixel_size + gap
+        gap = (asic_gap_h * unit, asic_gap_v * unit)
+        row_sz = cls.frag_ss_pixels * cls.pixel_size
+        col_sz = cls.frag_fs_pixels * cls.pixel_size
         for tileno in range(4):
             row, col = tileno // 2, tileno % 2
             tiles.append(GeometryFragment(
                 corner_pos=np.array(
-                    [col_sz - col * (col_sz + gap) - x0,
-                     row_sz - row * (row_sz + gap) - y0,
+                    [col_sz - col * (col_sz + gap[0]) - x0 + gap[0] / 2,
+                     row_sz - row * (row_sz + gap[1]) - y0 + gap[1] / 2,
                      0]),
                 ss_vec=np.array([0, -1, 0]) * cls.pixel_size,
                 fs_vec=np.array([-1, 0, 0]) * cls.pixel_size,
@@ -1810,13 +1815,15 @@ class Epix100Geometry(EpixGeometryBase):
     detector_type_name = 'ePix100'
     pixel_size = 50e-6
     inner_pixel_size = 175e-6
-    asic_gap = inner_pixel_size / pixel_size - 1
-    frag_ss_pixels = 352  # rows
+    asic_gap = 2 * (inner_pixel_size - pixel_size) / pixel_size
+    # ePix100 should have 352 pixel per row, but the data we store has 354
+    frag_ss_pixels = 354  # rows
     frag_fs_pixels = 384  # columns
     expected_data_shape = (
         EpixGeometryBase.n_modules,
         2 * frag_ss_pixels,
-        2 * frag_fs_pixels)
+        2 * frag_fs_pixels
+    )
 
 
 class Epix10KGeometry(EpixGeometryBase):
@@ -1837,10 +1844,11 @@ class Epix10KGeometry(EpixGeometryBase):
     detector_type_name = 'ePix10K'
     pixel_size = 100e-6
     inner_pixel_size = 250e-6
-    asic_gap = inner_pixel_size / pixel_size - 1
+    asic_gap = 2 * (inner_pixel_size - pixel_size) / pixel_size
     frag_ss_pixels = 176  # rows
     frag_fs_pixels = 192  # columns
     expected_data_shape = (
         EpixGeometryBase.n_modules,
         2 * frag_ss_pixels,
-        2 * frag_fs_pixels)
+        2 * frag_fs_pixels
+    )
--- a/extra_geom/tests/test_epix_geometry.py
+++ b/extra_geom/tests/test_epix_geometry.py
@@ -17,7 +17,7 @@ def args(request):
 @pytest.fixture
 def epix100():
     return (
-        Epix100Geometry.from_origin(), (352, 384), 50e-6,
+        Epix100Geometry.from_origin(), (354, 384), 50e-6,
         extra_geom.detectors.Epix100Geometry)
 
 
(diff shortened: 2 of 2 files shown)


extra_geom/detectors.py
```diff
@@ -1810,13 +1815,15 @@ class Epix100Geometry(EpixGeometryBase):
     detector_type_name = 'ePix100'
     pixel_size = 50e-6
     inner_pixel_size = 175e-6
-    asic_gap = inner_pixel_size / pixel_size - 1
-    frag_ss_pixels = 352  # rows
+    asic_gap = 2 * (inner_pixel_size - pixel_size) / pixel_size
+    # ePix100 should have 352 pixel per row, but the data we store has 354
+    frag_ss_pixels = 354  # rows
```
takluyver, line 1820: Can we confirm that this affects all ePix100 detectors we have/will have, and all data we store? I.e. it's not a quirk of one particular detector, or something that e.g. detector calibration might lop off. I don't want to fix the problem in one case and create a problem somewhere else.

As with my comment on #96, I think the priority is to understand exactly where the discrepancy comes from.

extra_geom/detectors.py
```diff
@@ -1682,19 +1682,24 @@ def from_origin(cls, origin=(0, 0), asic_gap=None, unit=None):
         if unit is None:
             unit = cls.pixel_size
         if asic_gap is None:
-            asic_gap = cls.asic_gap
+            asic_gap_h = asic_gap_v = cls.asic_gap
+        elif isinstance(asic_gap, Collection):
```
takluyver, line 1686: If we're going to be all precise about types here, I think we need a `Sequence` here. `Collection` covers things like sets and dicts as well.

extra_geom/detectors.py
```diff
@@ -1682,19 +1682,24 @@ def from_origin(cls, origin=(0, 0), asic_gap=None, unit=None):
...
+            asic_gap_h = asic_gap[0]
+            asic_gap_v = asic_gap[1]
```
takluyver, line 1688:
```suggestion
            asic_gap_h, asic_gap_v = asic_gap
```

This gives us free errors if you pass a longer sequence.

extra_geom/detectors.py
tmichela: I forwarded you an email with implementation details, in short it read 708 rows in the device implementation, so all ePix100 have the same data shape. As far as I can tell, the calibration does not change the data shape.

egorsobolev: Unfortunately, it can be only a temporary solution. At the end of last week experts from SLAC informed us that two alternative geometries exist (I've forwarded the email to you):

1) One module geometry: one monolithic sensor with 4 ASICs. This geometry does not have gaps, only larger pixels on the inner edges of ASICs.

2) Two modules geometry: two monolithic sensors with 2 ASICs each. Each sensor is in the guard ring.

In the first case, there can be no gap, since the manufacturing is very accurate (nm).

In the second case, it is not a pixel gap. Modules are assembled mechanically and can be located with a shift of columns and with a rotation, and not even in the same plane. It requires multi-module geometry.

egorsobolev: I see, you already know about two layouts

egorsobolev: These two rows you added make no sense. They cannot be calibrated, and shouldn't be assembled in the frame. They even don't have positions, since they aren't wired to the sensor. They must be excluded from the geometry.

egorsobolev:
Unfortunately, it can be only a temporary solution. At the end of last week experts from SLAC informed us that two alternative geometries exist (I've forwarded the email to you):

In the first layout, there can be no gap, since the manufacturing is very accurate (nm).

In the second layout, it is not a pixel gap. Modules are assembled mechanically and can be located with a shift of columns and with a rotation, and not even in the same plane. I suppose 2nd layout requires multi-module geometry.

extra_geom/detectors.py
takluyver: We could also check the array shape passed in to EXtra-geom, and if there are 708 rows, trim away the extra 4 to leave the real pixels, as a convenience feature. This would be similar to what we do for PNCCD to allow passing in the 2 modules as a single image.

egorsobolev: Yes, we could. But is it worth? I suppose that the assembling method will be able to cut off extra rows from provided data but methods that provide positions for assembling will return positions without extra rows. And what to do with `data_coords_to_positions`?

In principle, it could be a convention, but misleading even more.

Maybe we could add an additional method for data normalization? (it also can add dummy dimension) And add corresponding documentation and examples?

tmichela: > They even don't have positions, since they aren't wired to the sensor.

Thanks, I originally thought they could be the larger edge pixels, I've removed that now.

tmichela:
> I think it is rather a temporary solution.

In the first layout, there can be no gap, since the manufacturing is very accurate (nm).

In the second layout, it is not a pixel gap. Modules are assembled mechanically and can be located with a shift of columns and with a rotation, and not even in the same plane. I suppose 2nd layout requires multi-module geometry.

Of course, they are aligned but the precision of the mechanical assembly is not so good. I'm afraid the misalignment can be easily compared to the pixel size in any parameter.

Thanks for the details

I agree with the temporary solution.
This is to enable users to use this in the short term (next week).
AFAIK, people have used similar geometry with this detector layout, so they might be happy with this 'accuracy' level (?)

extra_geom/detectors.py
```diff
@@ -1810,13 +1853,34 @@ class Epix100Geometry(EpixGeometryBase):
...
+        automatic reshaping to adding the modules dimension.
+        """
+        if data.shape[-2:] == (2 * cls.frag_ss_pixels + 4, 2 * cls.frag_fs_pixels):
+            # EuXFEL data stored extra 2 row per ASIC used for diagnostic
+            #   - pixel max (the first and last rows in the pixel array)
+            #   - baseline (next rows)
+            # we removed them here as they do not contain data
+            data = np.concatenate((data[..., 1:cls.frag_ss_pixels+1, :],
+                                   data[..., -1-cls.frag_ss_pixels:-1, :]),
+                                  axis=-2)
```
egorsobolev, line 1881:
```suggestion
            data = data[...,2:-2,:]
```
There are two diagnostic rows on one edge of each ASIC. But ASICs look at each other in the data. Therefore, it turns out that these are the first two and the last two rows in the original data.

egorsobolev:
> AFAIK, people have used similar geometry with this detector layout, so they might be happy with this 'accuracy' level (?)

I'm sure they are happy with it as it is so much better than no gaps.

But it would be great to somehow encourage them to refine the geometry. I am afraid that with this solution, they will postpone the problem for a long time.

tmichela:
Ok, before I go forward, If I followed all parallel threads 🙃 there are (at least) 2 question:

1) gap geometry:
- Do we want to implement in this PR a `position_sensor` to handle this case instead of the current changes?

2) additional diagnostic rows:
- do we handle these automatically?
- do we add a convenient method to reshape the data?
- do we leave that to users?

takluyver:
If this is something we want ready for people to use next week, I think it's OK to focus on a `from_*` constructor that only takes a gap for now. You can always use the `.offset()` method to adjust the positions of tiles.

takluyver:
For Q2: I'd say handle them automatically. A shape of (708, 768) is probably only likely to happen if the data includes those extra rows, we know which ones they are, so it should never be wrong to trim them off when assembling an image.

egorsobolev:
No objections. Should we expose the method which trims the diagnostics row? I think it can be useful if the user uses pixel coordinates instead of assembling a frame (e.g. passes data to the pyFAI)

takluyver:
I could go either way on exposing the method. I think you still need to know a bit about what's going on with the extra rows (i.e. to even know that you need this method), and if you know about it, slicing an array is probably shorter and clearer than calling a method on a geometry object.

egorsobolev:
Agree. At the same time, one may forget details, how to exactly data should be normalized to use with positions provided by EXtra-geom.

tmichela:
Thanks both for the review!

PR #128, opened 2021-11-12
Add geom.rotate()

tmichela:
For refining geometry, it's useful to be able to move elements at different levels - whole detector, quadrant, module or tile. EXtra-geom should provide methods to make this easy.

```python3
import numpy as np
from extra_geom import LPD_1MGeometry
%matplotlib notebook


quad_pos = [(11.4, 299), (-11.5, 8), (254.5, -16), (278.5, 275)]
geom = LPD_1MGeometry.from_quad_positions(quad_pos)

angles = np.zeros((16, 16, 3))
center = np.zeros((16, 16, 3))

angles[:4, ...] = (0, 0, -30)
center[:4, :, :] = np.mean(
    [t.centre() for m in geom.modules[:4] for t in m], axis=0)

for n, mod in enumerate(geom.modules[4:8], start=4):
    mod_centre = np.mean([t.centre() for t in mod], axis=0)
    angles[n, ...] = (0, 0, (-1)**n * n*-5)
    center[n, ...] = mod_centre

for m, mod in enumerate(geom.modules[8:12], start=8):
    for t, tile in enumerate(mod):
        angles[m, t] = (0, 0, -(m-7)*t)
        center[m, t, ...] = tile.centre()

for m, mod in enumerate(geom.modules[12:], start=12):
    for t, tile in enumerate(mod):
        angles[m, t] = (-(m-7)*t, 0, 0)
        center[m, t, ...] = tile.centre()

geom.rotate(angles, center).inspect()
```

geom-rotate

### DIFF
--- a/extra_geom/base.py
+++ b/extra_geom/base.py
@@ -810,11 +810,11 @@ class DetectorGeometryBase:
         increasing looking toward the detector front plan, x increasing to
         the left, y increasing to the top. Positive rotations are clockwise.
 
-In other words:
+        In other words:
 
-- Positive x angles tilt the top edge of the detector backwards, away from the source
-- Positive y angles tilt the right-hand edge (looking along the beam) away from the source
-- Positive z angles turn the detector clockwise (looking along the beam)
+        - Positive x angles tilt the top edge of the detector backwards, away from the source
+        - Positive y angles tilt the right-hand edge (looking along the beam) away from the source
+        - Positive z angles turn the detector clockwise (looking along the beam)
 
         By default, this rotates all modules & tiles.
         Returns a new geometry object of the same type.
--- a/extra_geom/base.py
+++ b/extra_geom/base.py
@@ -799,7 +799,8 @@ class DetectorGeometryBase:
             ] for m, module in enumerate(self.modules)
         ])
 
-    def rotate(self, angles, center=None, modules=np.s_[:], tiles=np.s_[:]):
+    def rotate(self, angles, center=None, modules=np.s_[:], tiles=np.s_[:],
+               degrees=True):
         """Rotate around first pixel
 
         Parameters
@@ -819,6 +820,9 @@ class DetectorGeometryBase:
             Select modules to rotate; defaults to all modules.
         tiles: slice
             Select tiles to move within each module; defaults to all tiles.
+        degrees: bool
+            If True (default), angles are in degrees. If False, angles are in
+            radians.
         """
         rot = np.asarray(angles)
         if rot.shape[-1] != 3:
@@ -884,9 +888,12 @@ class DetectorGeometryBase:
             )
 
         @lru_cache()
-        def _rot(deg):
+        def _rot(rotations):
             # generate rotation matrix for the given angles
-            w, p, k = np.deg2rad(deg)
+            if degrees:
+                w, p, k = np.deg2rad(rotations)
+            else:
+                w, p, k = rotations
 
             rot_x = np.array([[1, 0, 0],
                               [0, np.cos(w), -np.sin(w)],
--- a/extra_geom/base.py
+++ b/extra_geom/base.py
@@ -55,10 +55,10 @@ class GeometryFragment:
         pos = self.corner_pos + shift
         return type(self)(pos, self.ss_vec, self.fs_vec, self.ss_pixels, self.fs_pixels)
 
-    def rotate(self, angles, center):
-        ss_vec = self.ss_vec @ angles
-        fs_vec = self.fs_vec @ angles
-        pos = (self.corner_pos - center) @ angles + center
+    def rotate(self, matrix, center):
+        ss_vec = self.ss_vec @ matrix
+        fs_vec = self.fs_vec @ matrix
+        pos = (self.corner_pos - center) @ matrix + center
         return type(self)(pos, ss_vec, fs_vec, self.ss_pixels, self.fs_pixels)
 
     def snap(self, px_shape):
@@ -803,6 +803,13 @@ class DetectorGeometryBase:
                degrees=True):
         """Rotate part or all of the detector, making a new geometry.
 
+        The rotation is defined by composition of rotations about the axes of
+        the coordinate system (https://en.wikipedia.org/wiki/Euler_angles),
+        i.e. a rotation around the z axis rotates the xy (detector) plan.
+        We use the right-hand rule, xy being the detector plan with z
+        increasing looking toward the detector front plan, x increasing to
+        the left, y increasing to the top. Positive rotations are clockwise.
+
         By default, this rotates all modules & tiles.
         Returns a new geometry object of the same type.
 
@@ -841,7 +848,11 @@ class DetectorGeometryBase:
             If True (default), angles are in degrees. If False, angles are in
             radians.
         """
-        rot = np.asarray(angles)
+        if degrees:
+            rot = np.deg2rad(angles)
+        else:
+            rot = np.asarray(angles)
+
         if rot.shape[-1] != 3:
             raise ValueError('Rotation angles must be 3d (x, y, z). '
                              f'Last dimension was {rot.shape[-1]}')
@@ -869,6 +880,10 @@ class DetectorGeometryBase:
                     center = np.mean([t.centre() for t in module], axis=0)
                     mods.append(np.array([center for _ in module[tiles]]))
                 refs[modules, tiles] = np.array(mods)
+            else:
+                # Single rotation, the reference is the detector center
+                # no shift required (refs is already initialized to 0)
+                pass
         else:
             center = np.asarray(center)
             if center.shape[:-1] == sel_refs.shape[:2]:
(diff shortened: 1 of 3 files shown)


takluyver:
Nice! Do we have a particular use case that requires this?

tmichela:
That will be needed for assembling/projecting multiple independent detectors (the HED use case if you remember), whenever I can (or someone else) work on it 😅 

There were also recently discussion about automated geometry refinement with @daniloefl
That may be useful for that as well

extra_geom/base.py
```diff
@@ -791,3 +798,135 @@ def offset(self, shift, *, modules=np.s_[:], tiles=np.s_[:]):
...
+                    [[t.centre() for t in m[tiles]]
+                    for m in self.modules[modules]]
+                )
+            elif rot.shape[:-1] == sel_rot.shape[:1] or modules != np.s_[:]:
+                # Per-module rotation
+                mods = []
+                for module in self.modules[modules]:
+                    center = np.mean([t.centre() for t in module], axis=0)
+                    mods.append(np.array([center for _ in module[tiles]]))
+                refs[modules, tiles] = np.array(mods)
```
takluyver, line 871: It took me a while to realise that we don't need a final `else:` here, because refs already contains zeros, which are the default when the whole detector is selected.

extra_geom/base.py
```diff
@@ -791,3 +798,135 @@ def offset(self, shift, *, modules=np.s_[:], tiles=np.s_[:]):
...
+                sel_refs[:] = center
+            elif center.shape[:-1] == sel_refs.shape[:1]:
+                # Per-module offsets - broadcast across tiles
+                sel_refs[:] = center[:, np.newaxis]
+            elif center.shape[:-1] == ():
+                # Single shift - broadcast across modules and tiles
+                sel_refs[:] = center
+            else:
+                raise ValueError(
+                    f"Got {center.shape[:-1]} coordinates. Expected either a "
```
takluyver, line 885:
```suggestion
                    f"Got {center.shape[:-1]} center coordinates. Expected either a "
```

extra_geom/base.py
```diff
@@ -791,3 +798,135 @@ def offset(self, shift, *, modules=np.s_[:], tiles=np.s_[:]):
...
+            # per-module rotations - broadcast across tiles
+            sel_rot[:] = rot[:, np.newaxis]
+        elif rot.shape[:-1] == ():
+            # single rotation - broadcast across modules and tiles
+            sel_rot[:] = rot
+        else:
+            raise ValueError(
+                f"Got {rot.shape[:-1]} coordinates. Expected either a single "
+                f"coordinate (), a coordinate per module {sel_rot.shape[:1]} "
+                f"or a coordinate per tile {sel_rot.shape[:2]}"
```
takluyver, line 904: I wouldn't call these coordinates - is there a term for a set of 3 angles for 3D rotation? Or do people with more of a maths background call them coordinates?

extra_geom/base.py
```diff
@@ -791,3 +798,135 @@ def offset(self, shift, *, modules=np.s_[:], tiles=np.s_[:]):
...
+            )
+
+        @lru_cache()
+        def _rot(rotations):
+            # generate rotation matrix for the given angles
+            if degrees:
+                w, p, k = np.deg2rad(rotations)
```
takluyver, line 911: I might do this conversion outside the inner function - it should be easy to apply it to the whole `rot` array, I think.

extra_geom/base.py
```diff
@@ -54,6 +55,12 @@ def offset(self, shift):
         pos = self.corner_pos + shift
         return type(self)(pos, self.ss_vec, self.fs_vec, self.ss_pixels, self.fs_pixels)
 
+    def rotate(self, angles, center):
```
takluyver, line 58: The `angles` parameter here should probably be called something like `matrix`, right?

extra_geom/base.py
```diff
@@ -791,3 +798,135 @@ def offset(self, shift, *, modules=np.s_[:], tiles=np.s_[:]):
...
+            geom2 = geom.rotate((0, 0, 90))
+
+            # Move the tile 0 in the module 0 around its center by 90 degrees
+            geom2 = geom.rotate((0, 0, 90), modules=np.s_[:1], tiles=np.s_[:1])
+
+            # Rotate each module by a separate amount
+            rotate = np.zeros((16, 3))
+            rotate[5] = (3, 5, 1)  # x, y, z for individual modules
+            rotate[10] = (0, -2, 1)
+            geom2 = geom.rotate(rotate)
```
takluyver, line 821: I'd like to have a brief description of what a rotation in each dimension means - maybe this is obvious to other people, but it always hurts my head that you specify an angle for z to rotate in the x-y plane (I understand why, it's just not intuitive).

In particular, I think it would be useful to have a reminder which way is 'forwards' for each dimension - e.g. is z rotation clockwise looking along the beam, or anticlockwise? Again, maybe this is second nature to some users, but probably not all.

extra_geom/base.py
tmichela: Want me to add an empty branch with that comment, for clarity?

extra_geom/tests/test_lpd_geometry.py
```diff
@@ -149,6 +149,100 @@ def test_offset():
...
+    y = np.array([m[0].corner_pos[1] for m in all_rotated.modules])
+    np.testing.assert_allclose(y, 0., atol=1e-7)
+
+    all_rotated = geom.rotate((0, 90, 0))
+    x = np.array([m[0].corner_pos[0] for m in all_rotated.modules])
+    np.testing.assert_allclose(x, 0., atol=1e-7)
+
+    all_rotated = geom.rotate((0, 0, 90))
+    assert all_rotated.output_array_for_position_fast().T.shape == \
+        geom.output_array_for_position_fast().shape
```
takluyver, line 167: Neat :laughing:

takluyver (commented):
Nice work. The actual mechanism looks good as far as I can see, my suggestions are just about clarity.

I'm not all that well up on the maths used here - it looks reasonable based on what I do know, but if you want that aspect reviewed carefully, we might want to ask someone else to look at it.

extra_geom/base.py
daniloefl: In analytical mechanics people may refer to angles as "generalized coordinates". For example, Euler angles or polar angles could be referred to as coordinates in this context (https://en.wikipedia.org/wiki/Generalized_coordinates). I would not see this as a problem. Even if it is, I think it is clear what is meant by context, right?

daniloefl:
Hi @tmichela , just thinking from my perspective, when trying to use this in the future, would people really specify directly the angles of the modules if they were to measure the geometry directly?

I just wonder if people wouldn't perhaps prefer to specify the x-y coordinates of the edges of each module instead (in case it is easier to measure them). If this is a more common thing to specify, perhaps a function could be added to convert that into angles. For my use case, this would not matter, but if people do not specify the angles directly, I would prefer to try to automatize whatever is more meaningful to people.

If people are indeed used to specify the angles, then never mind.

extra_geom/base.py
takluyver: Or just a comment where the `else` would go - whatever you think is clear.

extra_geom/base.py
takluyver: If you're passing both angles and centre coordinates, and you get an error talking about 'coordinates', I'd probably guess it meant the centre coordinates rather than the angles. It's probably not that hard to work out, I was just wondering if we could find a clearer wording.

tmichela:
@daniloefl Good point. Actually I don't know what would be easier/prefered for measurement. I think no-one has done that as far as I know. For "fixed" detectors, I would assume that they (modules to quadrant to detector to bench) *should* be all orthogonal to the beam axis and if there are misalignment, it should be too small to be easily measurable anyway 🤔 
For movable detectors, I hope we can (one day) rely on motors encoders to deduce spacial position and rotation directly (internal geometry of detectors in this case should be simpler, I believe).
In any case, I agree, adding a function to convert edges to angles makes sens 👍

extra_geom/base.py
tmichela: I changed `coordinates` to `rotation sequences`. Is it clearer? Or do you have another suggestion?

extra_geom/base.py
tmichela: I tried my best, let me know if that's clear enough 🙃

extra_geom/base.py
```diff
@@ -803,6 +803,13 @@ def rotate(self, angles, center=None, modules=np.s_[:], tiles=np.s_[:],
                degrees=True):
         """Rotate part or all of the detector, making a new geometry.
 
+        The rotation is defined by composition of rotations about the axes of
+        the coordinate system (https://en.wikipedia.org/wiki/Euler_angles),
+        i.e. a rotation around the z axis rotates the xy (detector) plan.
+        We use the right-hand rule, xy being the detector plan with z
+        increasing looking toward the detector front plan, x increasing to
+        the left, y increasing to the top. Positive rotations are clockwise.
+
```
takluyver, line 812:
```suggestion

In other words:

- Positive x angles tilt the top edge of the detector backwards, away from the source
- Positive y angles tilt the right-hand edge (looking along the beam) away from the source
- Positive z angles turn the detector clockwise (looking along the beam)

```

I think some concrete descriptions like this are useful alongside the definitions.

extra_geom/base.py
takluyver: Thanks, I think that's clear enough.

takluyver:
LGTM

tmichela:
Thanks for the review! :)

PR #139, opened 2022-02-21
Add .to_pyfai_detector() method

takluyver:
Convert EXtra-geom objects to PyFAI detector instances. Use like this:

```python
lpd_pyfai = geom.to_pyfai_detector()  # 

import pyFAI.azimuthalIntegrator

ai = pyFAI.azimuthalIntegrator.AzimuthalIntegrator(
    detector=lpd_pyfai,
    dist=0.2,
    wavelength=12.3984 / 9.3 * 1e-10
)
rint, I = ai.integrate1d(
    frame.reshape(16*256, 256),
    npt=300,
    unit="r_mm",
)
```

AGIPD, LPD, DSSC are done here. PyFAI partly supports DSSC's hexagonal pixels, and [full pixel-splitting support is planned](https://github.com/silx-kit/pyFAI/issues/1637).

It should in principle be easy to extend this to other detectors, but I wanted to make sure it works for these ones first. I also need to ask about how to handle detectors with variable numbers of modules, like Jungfrau.


tmichela:
nice, that's easier than I would have thought.

LGTM

takluyver:
I added several packages used in the tests to the `constraints.txt` file to control the version used by the CI. It's not all the packages that are getting installed - it's quite a long list because of all the Jupyter stuff nbval pulls in, and I couldn't be bothered to add them all, so I tried to focus on things we use directly, where changes are most likely to cause issues.

PR #142, opened 2022-02-26
Update faq.rst

JunCEEE:
The `imshow` of the image (at least for AGIPD) is not only upside down but also flipped horizontally.

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -6,7 +6,7 @@ Why is my image "flipped" when I display it?
 
 At EuXFEL, the convention is to show detector images looking at the front of the detector, meaning the beam coming from behind you. You can get the other view with EXtra-geom's plotting methods by passing ``frontview=False``.
 
-If you use matplotlib's ``imshow(array)`` to display an image, it may be upside down. You can correct this by passing ``imshow(array, origin='lower')``.
+If you use matplotlib's ``imshow(array)`` to display an image, it may be upside down. You can correct this by passing ``imshow(array, origin='lower')``. If you also want to flip the displayed image horizontally, pass ``arr[:, ::-1]`` or ``np.fliplr(arr)``.
 
 What are "slow-scan" and "fast-scan"?
 -------------------------------------


takluyver:
Thanks, but this is what the paragraph above is trying to explain. There isn't a 'right way' round for detector images horizontally, because it depends whether you imagine looking at the detector from the front, with the beam coming from behind you, or from the back, with the beam coming towards you.

JunCEEE:
Hi @takluyver, thanks for your prompt reply. You are right that there is not a only "right way". The PR doesn't intend to imply that there is only one correct way. 

The suggestion here is to remind the users that the image plotted by `imshow` could not only be upside down but also horizontally flipped. It also gives a quick solution for the users to flip them (many users have very limited python knowledge, and there is no `frontview` argument within the `position_modules_fast()` method.).

I created this PR based on my observation that some users are confused about which is the best way to get the offline analysis image in geometry the same as what they saw during the experiment, which is the geometry used for online data analysis. 

How about rewriting to:
> If you use matplotlib's ``imshow(array)`` to display an image, it may be upside down. You can correct this by passing ``imshow(array, origin='lower')``.  
The output of `position_modules_fast()` is in backview. You can view it in frontview by `imshow(array[:,::-1], origin='lower')` or `imshow(array[::-1,::-1])`.

Maybe add `frontview` argument to the `position_modules_fast()` method can make life easier?

takluyver:
How about adding something like this to the end of the paragraph about imshow:

> If you also want to flip the displayed image horizontally, pass `arr[:, ::-1]` or `np.fliplr(arr)`.

-----

> Maybe add frontview argument to the position_modules_fast() method can make life easier?

My instinct is to say no, because there are already two pretty good ways to flip the output horizontally: adding another one means more to understand. Also, if people want to display an image, I'd encourage them to use `.plot_data_fast()`, which also includes things like the centre marker and colour bar automatically.

But I'd be interested to hear what other people in DA think - @tmichela @philsmt ?

JunCEEE:
> How about adding something like this to the end of the paragraph about imshow:
> 
> > If you also want to flip the displayed image horizontally, pass `arr[:, ::-1]` or `np.fliplr(arr)`.

It looks good.


> > Maybe add frontview argument to the position_modules_fast() method can make life easier?
> 
> My instinct is to say no, because there are already two pretty good ways to flip the output horizontally: adding another one means more to understand. Also, if people want to display an image, I'd encourage them to use `.plot_data_fast()`, which also includes things like the centre marker and colour bar automatically.

About image plotting, from my SPI experience for AGIPD, we often need to choose an ROI to plot/analyze (like the [LDP example](https://extra-geom.readthedocs.io/en/latest/apply_geometry.html)). I don't know if there is a good way to `plot_data_fast()` with ROI, if so, we then don't need to bother to use the `position_modules_fast()` function.

takluyver:
> I don't know if there is a good way to plot_data_fast() with ROI

Good question, I hadn't considered this. There's a moderately good way, which looks like this:

```python
ax = geom.plot_data_fast(modules_data[12], vmin=0, vmax=3000)
ax.set_xlim(200, -200)
ax.set_ylim(-200, 200)
```

Here the numbers are in pixel units, but if you pass `axis_units='m'`, you can specify the limits in metres instead.

image

JunCEEE:
Thanks, @takluyver, it looks nice. That snippet would be helpful if it can be posted in the documentation.
I've edited this PR as you suggested. Now, the PR is at your disposal.

takluyver:
Since I was changing the relevant bit of docs in #143 already, I've added this example there.

PR #143, opened 2022-03-01
Drop _fast suffix on method names

takluyver:
For discussion:

The 'fast' in `position_modules_fast` and `plot_data_fast` seems kind of meaningless: there's a slower option implemented only for AGIPD, it's *really* slow and I think no-one uses it. And while it's faster than interpolation, it's not as fast as it could be (as @JamesWrigley found). So if we were designing this API from scratch, I'd leave `_fast` out of the names.

On the other hand, renaming things always adds a bit of mental effort. Of course we'd leave the old names as aliases so code doesn't break, but people will still notice that some code uses `position_modules` and some uses `position_modules_fast`, and wonder if there's a difference. I don't know if saving a few characters is worth that extra effort. :shrug:


philsmt:
I've always found it awkward those methods have a `_fast` suffix, especially without a corresponding non-fast one. It may also yet bring us in the unfortunate situation to at some point require a `_faster` method, or make the choice between a `_fast` version and any other suffix unintuitive.

I propose:
* Remove the `_fast` suffix from all current methods to define them as baseline
* Keep compatibility symbols for a while, possibly raising `DeprecationWarning`s
* Never use such a generic suffix again but specific ones (like `_lut` for @JamesWrigley's code)

takluyver:
> Keep compatibility symbols for a while, possibly raising DeprecationWarnings

Yup, definitely. I might leave them there almost indefinitely: it's very little effort to leave aliases around, and it always annoys people if you break their code.

> Never use such a generic suffix again but specific ones (like _lut for @JamesWrigley's code)

:+1: Another possible approach is to have a `method=` parameter (as PyFAI does). If the results are the same, though, we can just pick the fastest available.

philsmt:
> 👍 Another possible approach is to have a `method=` parameter (as PyFAI does). If the results are the same, though, we can just pick the fastest available.

In general yes, although I dislike this variant when arguments (even if only optional ones) diverge, which is not unlikely with a fancy assembly algorithm.

PR #160, opened 2022-05-19
Shift DSSC alternate rows in the opposite direction

takluyver:
@leguyader has found that we seem to have got the hexagon pattern wrong, and the odd-numbered rows should be shifted the other way. Data with sharp vertical edges looks much clearer with this change, when you plot each pixel as a hexagon (which I'll add a method for separately).

We're still checking this with detector experts, but it looks like we've got a mix of diagrams looking into vs. along the beam, and confusion over which pixel is (0, 0) led to this mistake.


CammilleCC:
L(!)GTM, I also learned from this :D

PR #163, opened 2022-05-30
FIX: avoid scientific formatting in crystfel file

tmichela:
`cfelpyutils` does parse scientific notation.


takluyver:
Thanks, LGTM

PR #167, opened 2022-06-15
Method to plot data on hexagons for DSSC

takluyver:
`geom.plot_data_hexes()` takes a stacked frame of DSSC data, and draws each pixel as a hexagon - the difference is only really visible when you zoom in. cc @leguyader, and thanks for the pointers to get started with this.

This is fairly slow, because matplotlib is processing 1M hexagons. I'm guessing it's not something that would be used that often, so performance isn't so important. But maybe we want to add an option to draw only a single module?

image

image

### DIFF
--- a/docs/geometry.rst
+++ b/docs/geometry.rst
@@ -220,6 +220,8 @@ approximately half a pixel width from their true position.
 
    .. automethod:: plot_data
 
+   .. automethod:: plot_data_hexes
+
    .. automethod:: position_modules
 
    .. automethod:: output_array_for_position
--- a/extra_geom/detectors.py
+++ b/extra_geom/detectors.py
@@ -1415,7 +1415,7 @@ class DSSC_1MGeometry(DetectorGeometryBase):
 
         collection = PolyCollection(
             [self._pixel_corners[::-1].T * self.pixel_size],
-            offsets=self.get_pixel_positions(centre=False)[..., :2].reshape(-1, 2),
+            offsets=px_offsets,
             transOffset=IdentityTransform(),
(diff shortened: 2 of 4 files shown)


leguyader:
I would support the possibility of plotting a single module.

But even in the case of a single module it is quite slow to plot hexagons. For that reasons I decided to keep plotting data as is on square lattice without geometry. So this is nice feature to have, but the induced slowness prevent this from being used more often.

takluyver:
Thanks @leguyader . I've added an option to plot a single module by specifying the module number. I think I managed to find a more efficient way of doing it based on the code in matplotlib's `hexbin()` function - it's still not the quickest plotting option, but hopefully for inspecting a single module it should be OK.

image

tmichela:
> image

It looks like the longer axis is the vertical one here, is that correct? (I always thought it was the other one... 🤔 )

Other than that this LGTM

takluyver:
I think this fits with the diagrams we've been given:

Pixel_orientation

tmichela:
Alright then 👍

leguyader:
> Thanks @leguyader . I've added an option to plot a single module by specifying the module number. I think I managed to find a more efficient way of doing it based on the code in matplotlib's `hexbin()` function - it's still not the quickest plotting option, but hopefully for inspecting a single module it should be OK.
> 
> image

If anything could we maybe make the colorbar horizontal for single module ? That would make a more compact figure with less empty white spaces

takluyver:
OK, I've done that and also changed the default figure size for a single module to take up less vertical space:

image

PR #169, opened 2022-06-29
FIX: JUNGFRAU position_modules with labelled array

tmichela:
It currently fails for labelled array because it expects `module` coortinates to start from `0`, but jungfrau has its modules labelled starting from `1`.

### DIFF
--- a/extra_geom/detectors.py
+++ b/extra_geom/detectors.py
@@ -8,7 +8,7 @@ import h5py
 import numpy as np
 
 from .base import DetectorGeometryBase, GeometryFragment
-
+from .snapped import isinstance_no_import
 
 class GenericGeometry(DetectorGeometryBase):
     """A generic detector layout based either on the CrystFEL geom file or on a set of parameters.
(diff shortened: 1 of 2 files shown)


extra_geom/detectors.py
```diff
@@ -1678,6 +1678,13 @@ def _tile_slice(cls, tileno):
         fs_slice = slice(tile_fs_offset, tile_fs_offset + cls.frag_fs_pixels)
         return ss_slice, fs_slice
 
+    def position_modules(self, data, out=None, threadpool=None):
+        if isinstance_no_import(data, 'xarray', 'DataArray'):
+            # we shift module indices by one as JUNGFRAU labels modules starting from 1..
+            # position_modules returns a numpy array so labels disapear anyway.
+            data['module'] = data['module'] - 1
```
takluyver, line 1685:
```suggestion
            data = data.copy(deep=False)
            data['module'] = data['module'] - 1
```

If I'm following [the docs](https://xarray.pydata.org/en/stable/generated/xarray.DataArray.copy.html) correctly, this should avoid modifying the object that's passed in, but without copying all the data inside it.

takluyver:
Other than that, LGTM, thanks :slightly_smiling_face:

extra_geom/detectors.py
tmichela: Good point, thanks!

PR #170, opened 2022-06-29
DRAFT: JUNGFRAU stripsel

tmichela:
This implements geometry for the JUNGFRAU stripsel module.

The implementation is slow and dirty, but I probably won't have much time to work on that in the next months, so that should allow working with it with the proper interface (and hopefully let someone with more time to improve it :) )

it's based on the requirements give on this [ticket](https://redmine.xfel.eu/issues/126444).
We don't have sample data yet for testing.

### DIFF
--- a/extra_geom/__init__.py
+++ b/extra_geom/__init__.py
@@ -35,8 +35,10 @@ program. If not, see <https://opensource.org/licenses/BSD-3-Clause>
 
 __version__ = "1.7.1"
 
-from .detectors import (AGIPD_1MGeometry, AGIPD_500K2GGeometry, DSSC_1MGeometry, Epix10KGeometry, Epix100Geometry,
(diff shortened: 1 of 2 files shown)


philsmt:
Note for later: All functions should accept both untransformed data in the regular shape as well as already transformed data without transforming it.

takluyver:
The reorganisation changes the array shape, doesn't it? So we can check the shape of a single module frame to see whether it's got already transformed data.

philsmt:
Yes, for the time being that should work out. I say `time being` as Marco hinted at further strixel-type sensors coming in with varying layouts, but this is all quite murky as of yet how they will distinguish themselves... we might ultimately require a generic pixel-reordering-machinery for (JF) detectors.

I tripped over this in the JF correct notebook, as naturally I want to reorder the pixels early after corrections, but then still plot it with EXtra-geom afterwards.

takluyver:
OK, so maybe we'll want an option to specify if the data is already transformed, and fall back to guessing based on the shape.

PR #174, opened 2022-08-11
DSSC cartesian conversion

takluyver:
Initial implementation of converting DSSC data to a cartesian grid (square pixels) and assembling it into a single image:

image

This relies on a small separate package which implements the conversion with Cython - [condat-gridconv](https://github.com/European-XFEL/condat-gridconv). Please look into that package as well as part of code review.

This converts each tile to a cartesian grid, but it still 'snaps' the tiles to pixel-aligned positions for assembly - there's no sub-pixel shifts when putting the tiles together. If that's desired, it will come at a later stage.

As shown in the screenshot above, this currently takes just over 1s to convert a 1 MPx frame. This is without any kind of parallelism - it should be pretty easy to speed it up by introducing some parallel processing. Of course, if you have many frames, you could parallelise it outside EXtra-geom as well.

@JamesWrigley @leguyader

To do:

- [x] Documentation for the new public methods
- [x] Tests


JamesWrigley (changes requested):
Code LGTM :+1: But, why not make `condat-gridconv` a required dependency? It doesn't have any dependencies other than numpy.

I think the code for `condat-gridconv` is also fine, but there's quite a few of my old comments still left in :sweat_smile: For example, this comment could be cleaned up to mention that we did end up using `-half_width`: https://github.com/European-XFEL/condat-gridconv/blob/main/src/condat_gridconv/gridconv.py#L85

takluyver:
Thanks James. Good point about the comments, I'll go through and try to do some cleanup.

My case for not making condat-gridconv a required dependency is that it should be straightforward to install EXtra-geom, preferably also on Mac/Windows and odd Linux environments (e.g. containers based on Alpine). That's all harder with a compiled extension unless you put in a bunch of work building wheels for different platforms (like NumPy, matplotlib and h5py do).

However, I hadn't spotted that cfelpyutils now builds an extension module as well (peakfinder8), so we've actually been requiring a compiled module with only Linux binaries since March (PR #114, released in 1.7). So I guess we have to decide whether to make cfelpyutils (used for reading .geom files) an optional dependency as well, or accept that it's a pain to set it up on anything besides Linux with glibc. :thinking:

JamesWrigley (approved):
LGTM!

PR #175, opened 2022-09-11
Fix AGIPD1M 'position_modules_interpolate', introduce upscale.

turkot:
I've fixed the 'position_modules_interpolate' method in AGIPD_1MGeometry class and introduced the 'scale' optional parameter in case someone would like to upscale. Part of the output array without scaling:
interpolate_scale_1
same part of the image with the scale of 4:
interpolate_scale_4
Of course scaling makes the interpolation slower.

### DIFF
--- a/extra_geom/detectors.py
+++ b/extra_geom/detectors.py
@@ -309,7 +309,7 @@ class AGIPD_1MGeometry(DetectorGeometryBase):
         ax.set_title('AGIPD-1M detector geometry ({})'.format(self.filename))
         return ax
 
-    def position_modules_interpolate(self, data):
+    def position_modules_interpolate(self, data, scale=1.0):
         """Assemble data from this detector according to where the pixels are.
 
         This performs interpolation, which is very slow.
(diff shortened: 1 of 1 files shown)


takluyver:
Thanks! Which bit was the actual fix? Is it the division by pixel size? If so, this has been broken for a long time, which probably illustrates that no-one has been using it.

If we're bringing this into use, and especially if we're introducing a new option that can make it even slower, I think we should try to improve the implementation so that it's significantly faster overall. Back when I wrote this, I made the odd decision to allocate a full layer for every tile (so a stack of 128 images of ~1 MPx each), transform each tile into one layer, and then use nanmax to reduce it down to a single image. I think the rationale was something about preserving NaN values in the gaps, but I don't think that's actually worth making it hideously slow for.

Maybe we can interpolate each tile in an array just large enough to hold it, and then copy it into the right place in the output image. Or maybe make a sparse array mapping input pixels to output pixels, so you flatten the input array and do a giant matrix multiplication to get the output? Or perhaps we just figure out how to take advantage of what PyFAI or some other package has already done.

turkot:
Yes, this and also the 'offset' was calculated in a weird way.
Yes, these sound like possible options. One other think we could do is to add an option to plot only a part (like roi) of the output image - for example, as in my plots shown before we don't need the whole detector image.

takluyver:
I think this is superseded by @tmichela's implementation in #395, so I'm going to close this PR.

PR #187, opened 2022-11-30
LPD Mini Geometry

cydanil:
This MR adds support for the LPD Mini Geometry.  

**Edit (TK)**: Our understanding of LPD Mini changed - see comments below. The original PR message is kept here:

-----

The LPD Mini is a new detector at FXE that is effectively equivalent to a one module of the LPD 1M.  
The advantages of the LPD Mini are its small physical size, allowing to move it, its panels that are interchangeable and can be positioned independently, and its new High-Z sensor technology.  

The addition proposed here is motivated by the need to create figures in the characterization and calibration reports, that use EXtra-geom.  
As such, my priority is to get a `LPD_MiniGeometry.inspect` function, although there's more to this  package, for which I will need your guidance: 
- I took inspiration of the `AGIPD_500K2GGeometry`, should the tests follow more the AGIPD500k or LPD ones?
- I think I should expand the example LPD rather than add a new one, as is done for the AGIPD?

Here's a screenshot of the geometry as a figure:

image


takluyver:
Nice, thanks!

I'm surprised the figure from `inspect()` doesn't look more square - the mini is 256 x 256 pixels, right? Do we need to do something to set the aspect ratio to make it a better shape by default?

There are three methods which we should implement to define how the geometry fits together - they're all pretty small, and the implementation can probably be copied from `LPD_1MGeometry`. These are `split_tiles`, `_tile_slice` and `_module_coords_to_tile`.

> I took inspiration of the AGIPD_500K2GGeometry, should the tests follow more the AGIPD500k or LPD ones?

What's tested for each detector is a bit ad-hoc - most of the actual code is shared between detectors, so when we've added new functionality, we often pick an arbitrary detector to write the tests for. So long as the tests broadly exercise the new code you've added, I'm happy.

> I think I should expand the example LPD rather than add a new one, as is done for the AGIPD?

Yeah, I think it makes sense to have an 'LPD geometry' notebook with both 1M and mini, rather than splitting those into separate notebooks. :+1:

tmichela:
Since this is essentially a LPD module, probably just overwriting LPD1M class properties should just work, something like:

```python
class LPD_64kGeometry(LPD_1MGeometry):
    detector_type_name = 'LPD-64k'
    pixel_size = 5e-4  # 5e-4 metres == 0.5 mm
    frag_ss_pixels = 32
    frag_fs_pixels = 128
    n_quads = 1
    n_modules = 1
    n_tiles_per_module = 16
    expected_data_shape = (1, 256, 256)
    _draw_first_px_on_tile = 8  # The first pixel in stored data is on tile 8
    _pyfai_cls_name = 'LPD64k'
```

Maybe one needs to change calculation of the 'center' and overwrite a few docstrings, but that should *probably* just work.

cydanil:
> Since this is essentially a LPD module, probably just overwriting LPD1M class properties should just work, something like:

This works pretty well for my needs, actually.  
I had not gone this route to mimic the implementation of AGIPD500k which is its own thing.  

I think it should have a `from_origin` function, as there's no assembly of quadrants, as for AGIPD500k or EpixBase.  
Does that seem right to you?

takluyver:
> I think it should have a from_origin function, as there's no assembly of quadrants, as for AGIPD500k or EpixBase.
Does that seem right to you?

I think so. And if we're deriving from LPD1M, we should probably also override some of the methods like `from_quad_positions` to raise an error.

Or I suppose a more proper way to do it would be to have a parent class for LPD geometry, and then two subclasses for LPD 1M and LPD mini.

cydanil:
I went with subclassing the LPD1M class, because it was fast and enabled me to add tests (based off the LPD1M tests).  

I will next update the docs and then see about creating a base LPD class, which would be cleaner as there's many `NotImplementedErrors` that needn't be there.

cydanil:
Thanks for your feedback above.  
I went back and implemented an `LPDGeometryBase` which is used by LPD1M and LPDMini in order to clean all the NotImplemented methods.  

I also updated the documentation, both in the rst and the LPD example notebook.  
I tested the documentation by going through each of the documented functions to ensure it's correct.  

I think it's ready, and would welcome your review.  
Cheers,  
Cyril

JamesWrigley (approved):
LGTM, but IANAE.

takluyver:
There's a difference between the inspect plot in the notebook and the one in the reference docs (the first pixel & first row parts):

image

image

I'm guessing one was generated before some tweak was made?

extra_geom/detectors.py
```diff
@@ -608,7 +608,81 @@ def agipd_asic_seams():
...
+class LPDGeometryBase(DetectorGeometryBase):
+    """Base class for LPD detector geometry. Subclassed for specific detectors.
+
+    The coordinates used in this class are 3D (x, y, z), and represent metres.
+    """
+    pixel_size = 5e-4  # 5e-4 metres == 0.5 mm
+    frag_ss_pixels = 32
+    frag_fs_pixels = 128
+    n_tiles_per_module = 16
+    _draw_first_px_on_tile = 9  # The first pixel in stored data is on tile 8
```
takluyver, line 620: I think this should be 8? It was before.

cydanil:
Thanks for the feedback, I was hoping to get away with not understanding what I'm doing, but it didn't work out.  

I fixed the geometry, and made the documentation consistent with the LPD 1M behaviour (top-left corner as reference).

extra_geom/detectors.py
```diff
@@ -930,67 +1000,102 @@ def inspect(self, axis_units='px', frontview=True):
...
+    The coordinates used in this class are 3D (x, y, z), and represent metres.
+
+    You won't normally instantiate this class directly:
+    use one of the constructor class methods to create or load a geometry.
+    """
+    detector_type_name = 'LPD-Mini'
+    n_quads = 1
+    n_modules = 1
+    pixel_size = 5e-4  # 5e-4 metres == 0.5 mm
+    expected_data_shape = (1, 256, 256)
```
takluyver, line 1016: I guess we might want to allow the LPD Mini class to handle 2D input arrays. We already do something similar for ePix - have a look at the `_ensure_shape` method.

extra_geom/detectors.py
```diff
@@ -930,67 +1000,102 @@ def inspect(self, axis_units='px', frontview=True):
...
+    _pyfai_cls_name = 'LPDMINI'
 
     @classmethod
-    def _module_coords_to_tile(cls, slow_scan, fast_scan):
-        tiles_across, tile_fs = np.divmod(fast_scan, cls.frag_fs_pixels)
-        tiles_up, tile_ss = np.divmod(slow_scan, cls.frag_ss_pixels)
+    def from_origin(cls, origin=(0, 0), asic_gap=4, unit=pixel_size):
```
takluyver, line 1020:
```suggestion
    def from_origin(cls, origin=(0, 0), asic_gap=4, unit=1e-3):
```

I think we should be consistent with LPD-1M's `from_quad_positions` here, rather than the `from_origin` method for AGIPD-500k. It's unfortunate that we've ended up with different default units for different detectors, but I'd like to aim for consistency within each detector family.

(The docstring will need a change too)

extra_geom/tests/test_lpdmini_geometry.py
```diff
@@ -0,0 +1,200 @@
...
+    src_file = pjoin(tests_dir, 'lpd_mar_18.h5')
+    dst_file = pjoin(str(tmpdir), 'lpd_inverted.h5')
+    invert_xfel_lpd_geom(src_file, dst_file)
+    with h5py.File(src_file, 'r') as fsrc, h5py.File(dst_file, 'r') as fdst:
+        np.testing.assert_array_equal(
+            fsrc['Q1/M1/Position'][:], -1 * fdst['Q1/M1/Position'][:]
+        )
+        np.testing.assert_array_equal(
+            fsrc['Q1/M1/T07/Position'][:], -1 * fdst['Q1/M1/T07/Position'][:]
+        )
```
takluyver, line 194: This is directly copied from the LPD 1M tests, right? Probably no need to repeat it.

takluyver:
Just to note here, this will need an update, because we've discovered that an LPD mini is actually a pair of 32 x 128 tiles, arranged in a 64 x 128 module (but with the data stored as 32 x 256, unless the DAQ reshapes it). LPD are apparently planning to use two of these mini modules together.

takluyver:
It took longer than I expected, but I've updated this now that we better understand what LPD Mini is. Namely, a variable number of 2-tile modules (64 x 128 pixels, but they will be represented in the data as 32 x 256 for reasons).

image

One tile is rotated 180 degrees with regard to the other, so the coordinates go in opposite directions. I'm not actually sure which direction should be which horizontally - for now I've assumed that the diagrams in the documentation we've been given are looking 'along the beam', but if they're 'into the beam' then we'll need to flip everything round.

I've assumed that the input data for assembly will have the 32 x 256 slices for each relevant module cut out before it's passed to EXtra-geom. We can make an EXtra-data component to help with this.

extra_geom/detectors.py
```diff
@@ -993,6 +999,151 @@ def to_distortion_array(self, allow_negative_xy=False):
...
+    expected_data_shape = (0, 32, 256)
+    _pyfai_cls_name = 'LPDMINI'
+
+    def __init__(self, modules, filename='No file', metadata=None):
+        super().__init__(modules, filename, metadata)
+        self.expected_data_shape = (len(modules), 32, 256)
+        self.n_modules = len(modules)
+
+    @classmethod
+    def from_module_positions(cls, positions, rotations=None, asic_gap=None, unit=1e-3):
```
takluyver, line 1022: Now that I've written the support for in-plane rotations (which did involve some swearing getting all the signs right)... is it actually necessary? We could lay the modules out without rotation, and then use `.rotate()` afterwards.

I was prompted by a diagram showing that a planned use case has two modules turned on their side and stacked together vertically (making a column of 64 px width by 256 pixels height).

I'm inclined to leave it in, because to define the geometry with the `rotate()` method, you'd either have to figure out fake corner positions so that they rotate into the right places, or fiddle with specifying the corner position of each module as the centre for its rotation.

We could also come up with some way of specifying it that only allows rotations in multiples of 90 degrees, to simplify things a bit, but now that I've (hopefully) got it working, I don't think there's much advantage to that.

cydanil (commented):
It seems that the figure in `docs/lpd_geometry.ipynb` needs to be updated (ie. the notebook reran).

takluyver:
> It seems that the figure in docs/lpd_geometry.ipynb needs to be updated (ie. the notebook reran).

I just re-ran the notebook locally and nothing changed. What do you see wrong with it? :confused: 

image

image

cydanil:
Thank you @takluyver for seeing this work to its completion, LGTM.

takluyver:
Thank-you! :smile:

PR #190, opened 2022-12-05
Fix pyfai lpd alias

cydanil:
This MR fixes a typo found in the LPD 1M PyFAI exports.  

Thanks,
Cyril


takluyver:
:facepalm: Thanks!

PR #197, opened 2023-01-17
Implement 'to_pyfai_detector' for JUNGFRAU detectors geometry.

turkot:
Implementation of 'to_pyfai_detector' method for  JUNGFRAU detectors geometry.

### DIFF
--- a/extra_geom/base.py
+++ b/extra_geom/base.py
@@ -659,7 +659,10 @@ class DetectorGeometryBase:
             raise NotImplementedError
 
         from . import pyfai
-        det = getattr(pyfai, self._pyfai_cls_name)(n_modules=self.n_modules)
+        if 'JUNGFRAU' in self.detector_type_name:
+            det = getattr(pyfai, self._pyfai_cls_name)(n_modules=self.n_modules)
+        else:
+            det = getattr(pyfai, self._pyfai_cls_name)()
         det.set_pixel_corners(self.to_distortion_array(allow_negative_xy=True))
         return det
 
--- a/extra_geom/base.py
+++ b/extra_geom/base.py
@@ -659,10 +659,7 @@ class DetectorGeometryBase:
             raise NotImplementedError
 
         from . import pyfai
-        if 'JUNGFRAU' in self.detector_type_name:
-            det = getattr(pyfai, self._pyfai_cls_name)(n_modules=self.n_modules)
-        else:
-            det = getattr(pyfai, self._pyfai_cls_name)()
+        det = getattr(pyfai, self._pyfai_cls_name)()
         det.set_pixel_corners(self.to_distortion_array(allow_negative_xy=True))
         return det
 
--- a/extra_geom/base.py
+++ b/extra_geom/base.py
@@ -659,7 +659,7 @@ class DetectorGeometryBase:
             raise NotImplementedError
 
         from . import pyfai
-        det = getattr(pyfai, self._pyfai_cls_name)()
+        det = getattr(pyfai, self._pyfai_cls_name)(n_modules=self.n_modules)
         det.set_pixel_corners(self.to_distortion_array(allow_negative_xy=True))
         return det
 
--- a/extra_geom/pyfai.py
+++ b/extra_geom/pyfai.py
@@ -46,7 +46,6 @@ class LPD1M(Detector):
 
 class JUNGFRAU_EuXFEL(Detector):
     IS_CONTIGUOUS = False
-    MAX_SHAPE = (0, 256)  # num modules filled at instantiation
 
     def __init__(self, pixel1=7.5e-5, pixel2=7.5e-5, n_modules=None, **kwargs):
         self.MAX_SHAPE = (n_modules*256, 256)
--- a/extra_geom/pyfai.py
+++ b/extra_geom/pyfai.py
@@ -44,10 +44,9 @@ class LPD1M(Detector):
         super().__init__(pixel1, pixel2, **kwargs)
 
 
-class JUNGFRAU(Detector):
+class JUNGFRAU_EuXFEL(Detector):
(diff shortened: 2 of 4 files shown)


takluyver:
PyFAI already has some classes for JUNGFRAU with specific numbers of modules (500k, 4M, 16M), which we may be able to use:

https://github.com/silx-kit/pyFAI/blob/main/pyFAI/detectors/_psi.py

@kif do you have an idiomatic way to handle detectors with varying numbers of modules? I know we've got at least a JUNGFRAU 1M installation (2 modules) at XFEL in addition to single modules and 4M.

takluyver:
Thanks Jérôme!

> I am a bit surprised since the module in our Jungfrau is 512x1024 (with gaps between sub-modules which are 256x256). Your construction with just submodules looks over-complicated to me since all 8 sub-modules of a module are attached to the same Si-wafer and have been manufactured within the same ASIC (IIRC), thus there should not be any positioning issue between them.

That's just how our code has grown. It's more complicated in one way, because we don't need to position these sub-modules independently, but simpler in another, because there's no need for a separate step to account for the gaps. We always save and stream JUNGFRAU data as (512, 1024) arrays without the gaps.

Given that difference, maybe it's simpler for us to define our own class(es) for JUNGFRAU at EuXFEL. :slightly_smiling_face: 

> Considering families of detectors, there are some in the _dectris module: the parent class contains the method to calculate the mask, then each detector from the family just defines the size

Thanks, I see the Eiger family there, for instance.

Do you foresee any issues with setting `MAX_SHAPE` as an instance attribute, as Oleksii has in this PR, rather than making lots of separate subclasses? We can do the subclasses if needed, but if the instance attribute works, that's pretty convenient for our use case.

> Pay attention about name collision when creating detectors since there is a metaclass to register all detectors and to build a factory. I have no idea how this would behave in case of name collision.

Good point, thanks. Maybe we can call it JungfrauEuxfel or something to avoid collisions. I think all the detectors we've previously defined are custom made for us, so this is the first time there's a  risk of a collision.

This applies to any aliases we define as well, right?

kif:
> Do you foresee any issues with setting `MAX_SHAPE` as an instance attribute, as Oleksii has in this PR, rather than making lots of separate subclasses? We can do the subclasses if needed, but if the instance attribute works, that's pretty convenient for our use case.

Should not be an issue.
Try copy.copy and other tools to duplicate a detector especially the detctor.factory.

> > Pay attention about name collision when creating detectors since there is a metaclass to register all detectors and to build a factory. I have no idea how this would behave in case of name collision.  
> 
> Good point, thanks. Maybe we can call it JungfrauEuxfel or something to avoid collisions. I think all the detectors we've previously defined are custom made for us, so this is the first time there's a  risk of a collision.
> 
> This applies to any aliases we define as well, right?

yep.

extra_geom/base.py
```diff
@@ -659,7 +659,10 @@ def to_pyfai_detector(self):
             raise NotImplementedError
 
         from . import pyfai
-        det = getattr(pyfai, self._pyfai_cls_name)()
+        if 'JUNGFRAU' in self.detector_type_name:
+            det = getattr(pyfai, self._pyfai_cls_name)(n_modules=self.n_modules)
```
takluyver, line 663: In general, the way to have special behaviour for one subclass like this is to override the method in that subclass, rather than putting an `if` in the base class. The idea is that the base class doesn't 'know' about its subclasses.

extra_geom/pyfai.py
```diff
@@ -42,3 +42,13 @@ class LPD1M(Detector):
 
     def __init__(self, pixel1=5e-4, pixel2=5e-4, **kwargs):
         super().__init__(pixel1, pixel2, **kwargs)
+
+
+class JUNGFRAU(Detector):
```
takluyver, line 47: From the discussion with Jérôme, we should call this something like JUNGFRAU_EuXFEFL to ensure we don't clash with the names already in PyFAI's class registry (and remove the alias).

extra_geom/pyfai.py
```diff
@@ -42,3 +42,13 @@ class LPD1M(Detector):
...
+    IS_CONTIGUOUS = False
+    MAX_SHAPE = (0, 256)  # num modules filled at instantiation
```
takluyver, line 49: If it works without this line, I'd get rid of it, so we don't have some deliberatly incorrect information.

extra_geom/pyfai.py
```diff
@@ -42,3 +42,13 @@ class LPD1M(Detector):
...
+    aliases = ["JUNGFRAU"]
+
+    def __init__(self, pixel1=7.5e-5, pixel2=7.5e-5, n_modules=None, **kwargs):
+        self.MAX_SHAPE = (n_modules*256, 256)
```
takluyver, line 53: Aren't JUNGFRAU modules 512 x 1024?

```suggestion
        self.MAX_SHAPE = (n_modules*512, 1024)
```

extra_geom/base.py
turkot: Thank you, implemented in https://github.com/European-XFEL/EXtra-geom/pull/197/commits/7a3d83e882ced59524f527dca74dac1249b8ed84.

extra_geom/pyfai.py
turkot: Thanks, renamed in https://github.com/European-XFEL/EXtra-geom/pull/197/commits/77df7fc01730ecb3fc36e78e624f64b43f3e9dcd.

extra_geom/pyfai.py
turkot: Sure, removed in https://github.com/European-XFEL/EXtra-geom/pull/197/commits/52b84b06891d7961558c542c16ea12b9a95b5418.

takluyver:
Thanks @turkot, this LGTM.

PR #198, opened 2023-01-18
Overwrite plot_data method to modify modules labels for Jungfrau

kakhahmed:
`plot_data()` for JUNGFRAUGeometry modifying the modules labels in an exact way done for `position_modules()` as Jungfrau modules start from 1

This PR is used and tested at


takluyver:
Thanks @kakhahmed, this LGTM. I'll go through a couple of the other open PRs and then see about a release.

PR #206, opened 2023-02-16
Set `module` coords to 0 for jungfrau detector of one module.

kakhahmed:
This MR fixes an issue when plotting xarray for FXE_XAD_JF500K which has one module `JNGFR03`.

This fix is used by https://git.xfel.eu/calibration/pycalibration/-/merge_requests/775

### DIFF
--- a/extra_geom/detectors.py
+++ b/extra_geom/detectors.py
@@ -1783,10 +1783,17 @@ class JUNGFRAUGeometry(DetectorGeometryBase):
         ax=None, figsize=None, colorbar=True, **kwargs,
     ):
         if isinstance_no_import(data, 'xarray', 'DataArray'):
+            # Jungfrau detector can have a module with a label higher than 1.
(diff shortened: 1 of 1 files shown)


takluyver:
I'm not sure this is quite right: if they called the solo module '1', would the 2 modules of JF1M be '2' and '3'?

We should check if any other instruments have done strange bits of numbering like this - if not, maybe the best thing is to special-case it in the EXtra-data component for the name `FXE_XAD_JF500K`.

kakhahmed:
> if they called the solo module '1', would the 2 modules of JF1M be '2' and '3'

Yes. So the same detector-type modules have an incrementing module number. For example, we can have the same issue with AGIPD if there is another added AGIPD at MID or SPB. The module numbers will start from 16,...


But what is the assumption this code is based on? That we can have two Jungfraus in the same instrument using JNFR01?

Adding a special case should work I guess, But I don't think it's a long-term solution, unfortunately.

PR #207, opened 2023-02-20
Fix DSSC cartesian assembly from xarray labelled arrays

### LINKED ISSUE #204: DSSC position_modules_cartesian() doesn't accept xarray object
@leguyader let me know that the DSSC method `.position_modules_cartesian()` doesn't accept an xarray DataArray, unlike the regular `position_modules` method, because it calls `.reshape()`, a method which these arrays don't have.

takluyver:
The other assembly methods accept a labelled array, so this one should too.

Closes #204.


CammilleCC:
LGTM

PR #213, opened 2023-03-27
FIX: AGIPD_500K2GGeometry.from_origin with unit position in meter

tmichela:
`.from_origin` was broken if the `unit` argument was different from the detector's pixel size.

### DIFF
--- a/extra_geom/detectors.py
+++ b/extra_geom/detectors.py
@@ -469,8 +469,10 @@ class AGIPD_500K2GGeometry(DetectorGeometryBase):
         To give positions in units other than pixels, pass the *unit* parameter
         as the length of the unit in metres. E.g. ``unit=1e-3`` means the
         coordinates are in millimetres.
+        *unit* only applies to the *origin* argument, *asic_gap* and *panel_gap*
+        are always given in pixel.
         """
-        asic_gap_px = asic_gap * unit / cls.pixel_size
+        origin = np.asarray(origin) * unit
         panel_gap_x = panel_gap[0] * cls.pixel_size
         panel_gap_y = panel_gap[1] * cls.pixel_size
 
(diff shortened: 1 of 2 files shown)


extra_geom/detectors.py
```diff
@@ -469,8 +469,10 @@ def from_origin(cls, origin=(0, 0), asic_gap=2, panel_gap=(16, 30),
         To give positions in units other than pixels, pass the *unit* parameter
         as the length of the unit in metres. E.g. ``unit=1e-3`` means the
         coordinates are in millimetres.
+        *unit* only applies to the *origin* argument, *asic_gap* and *panel_gap*
+        are always given in pixel.
```
takluyver, line 473: This would make it inconsistent with other detectors (AGIPD-1M, LPD, DSSC, JUNGFRAU, ePix all take a gap argument which unit does affect).

Of course that make a challenge for the default - if you only override the unit, it shouldn't change the default gap size. That's an outstanding issue for most detector classes, but LPD & DSSC have it fixed by using `None` as the default in the method and doing this:

https://github.com/European-XFEL/EXtra-geom/blob/874459af10dfb84bd0c8ffcb88f31c0014953ff7/extra_geom/detectors.py#L660-L662

My leaning would be to follow that strategy, keeping compatibility for anyone specifying both `unit` and gap sizes. But that's probably very few people, if any, so we could likely get away with saying that gap sizes are always in a fixed unit and not affected by the `unit` parameter. I'd like to keep it consistent either way, though.

extra_geom/detectors.py
```diff
@@ -454,23 +454,30 @@ class AGIPD_500K2GGeometry(DetectorGeometryBase):
...
+
+        To give positions in units other than pixels, pass the *unit* parameter as the
+        length of the unit in metres. E.g. ``unit=1e-3`` means the coordinates are in
+        millimetres.
         """
-        asic_gap_px = asic_gap * unit / cls.pixel_size
+        origin = np.asarray(origin) * unit
+        px_conversion = unit / cls.pixel_size
+        asic_gap = 2 if (asic_gap is None) else asic_gap * px_conversion 
+        panel_gap = (16, 30) if (panel_gap is None) else panel_gap * px_conversion
```
takluyver, line 479:
```suggestion
        panel_gap = (16, 30) if (panel_gap is None) else np.asarray(panel_gap) * px_conversion 
```

takluyver:
One more small comment, otherwise this LGTM, thanks

tmichela:
Thanks!

PR #226, opened 2023-06-01
Label AGIPD-500k with M0-M7 module numbers

### LINKED ISSUE #223: AGIPD500K does not have the channel numbers shown in `inspect()`

takluyver:
@tmichela @philsmt @daniloefl - does this look right? Do you think anyone will be upset to lose the QxMy numbering?

image

Closes #223.

### DIFF
--- a/extra_geom/detectors.py
+++ b/extra_geom/detectors.py
@@ -538,9 +538,8 @@ class AGIPD_500K2GGeometry(DetectorGeometryBase):
 
         # Label modules and tiles
         for ch, module in enumerate(self.modules):
-            quad, mod = (1 + v for v in divmod(ch, 4))
             cx, cy, _ = module[4].centre() * scale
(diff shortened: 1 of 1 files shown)


tmichela:
Is the `QxMy` labeling actually used beyond the calibration services?

takluyver:
For other detectors, the QxMy pattern is certainly used - e.g. see these docs for AGIPD at MID: https://rtd.xfel.eu/docs/agipd-manual-mid/en/latest/introduction.html#naming-convention

Whether it's specifically used for AGIPD 500k is the question at hand. :slightly_smiling_face:

tmichela:
Probably not, but who knows 🤷 
To be safe, I'd maybe include both notation (`QxMy (CHz)`) but also that's probably not too important, so that's up to you.

takluyver:
If we don't know of anyone using the QxMy notation for AGIPD 500k, and it does look like the Mx notation is used (I looked in Karabo), let's try this and see if anyone complains.

PR #243, opened 2023-08-31
Implement example geometries for all detectors

JamesWrigley:
We discussed this ages ago on Zulip (see the `Default geometries?` topic) but never got around to implementing it. I think most of the changes here are fairly straightforward, except for the LPD mini and JF which can have a variable number of modules. Here's what they look like:
image
image

The odd thing compared to the examples is that the origin is not centered. I tried centering it but got very confused and gave up :shrug: Geometry is quite bamboozling...

Some other things:
- You could argue that `PNCCDGeometry` and the epix classes don't need a `.example()` because some of their existing methods already have reasonable defaults. I'd say that it's good to have `.example()` methods anyway for consistency, but I don't feel very strongly about it.
- Going a bit further than `.example()`, the epix classes now have 'real' geometry methods `.monolithic_geometry()` and `.pair_geometry()` for convenience.

### DIFF
--- a/.github/dependabot/constraints.txt
+++ b/.github/dependabot/constraints.txt
@@ -7,6 +7,7 @@ h5py<3.10.0
 kiwisolver==1.4.5
 matplotlib<3.8.0
 nbval==0.10.0
+numexpr==2.8.4
 numpy<1.26.0
 Pillow<10.1.0
 pyFAI==2023.5.0
--- a/docs/geometry.rst
+++ b/docs/geometry.rst
@@ -41,6 +41,8 @@ but this geometry code works with a position for each tile.
 
    .. automethod:: from_crystfel_geom
 
+   .. automethod:: example
+
    .. automethod:: offset
 
    .. automethod:: rotate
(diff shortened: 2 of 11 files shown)


tmichela (commented):
Generally looks good to me, but I'd wait for @takluyver to be back and have a look at it.

I'd maybe also move the origin to a corner for detectors that have a rigid layout (no moving quadrant)

JamesWrigley:
Partially done in 370ec3b, new geometry:
image

image

I didn't do it for detectors like the PNCCD and epix's for aesthetics :stuck_out_tongue:

takluyver:
Thanks, I think this is a good idea. I agree with the change you've already made to put the origin at a corner for some of the less regular detectors.

For LPD mini, let's just put the N modules in a single column, rather than two columns. I think they've only got about 3 modules for this, and one of those is a different type of sensor, so you can't exactly use them all together as 'one detector'.

JamesWrigley:
Gotcha, fixed in b1190eb.

JamesWrigley:
Hmm, we seem to have run into this in CI: https://github.com/silx-kit/pyFAI/issues/1947
FWIW, all the tests pass locally for me.

takluyver:
It sounds like we can probably work around it by pinning numexpr in the [constraints file](https://github.com/European-XFEL/EXtra-geom/blob/master/.github/dependabot/constraints.txt).

takluyver:
Thanks, LGTM (once CI is sorted)

PR #269, opened 2024-02-05
Geometry update by motor positions of quadrants

egorsobolev:
This implements the update of geometry according to the motors of detector quadrants. The motor position can be write to and read from Crystfel geometry files (write as comments). There is also the helper to read motors from experimental data.

@takluyver @philsmt @JamesWrigley


takluyver:
Hi Egor, sorry it has taken me a while to look at this. Before I get into commenting on specific bits of the code, I have a few broader questions & thoughts.

Am I understanding correctly that we can only calculate quadrant shifts based on motor movements, not absolute positions? I.e. we can't translate motor positions directly to quadrant positions, we need a previous geometry and the difference in motor positions between the two?

I think the code to read motor positions from saved data (what's currently the `read_motors_from_data()` function) would be a more natural fit as an EXtra component which knows about the source name patterns - this is quite similar to the pattern of components we already have or have planned. It could look something like this:

```python
from extra.components import AGIPDQuadrantMotors

quad_motors = AGIPDQuadrantMotors(run)
quad_motors.get_positions(atol=0.01)
```

I don't particularly like the idea of storing motor positions in comments in .geom files - both because it's essentially defining a new ad-hoc file format, and also because if someone modifies the geometry with a tool that ignores the motor positions, they can be out of sync with the module positions they're stored with. But I also don't really have a better idea for how to store these at the moment.

egorsobolev:
> Am I understanding correctly that we can only calculate quadrant shifts based on motor movements, not absolute positions? I.e. we can't translate motor positions directly to quadrant positions, we need a previous geometry and the difference in motor positions between the two?

Yes. But its generalization. If you wanna use absolute position, you just define the reference quadrant positions for zero motor positions:
```python
    quad_pos_for_zero_motor_pos = [(-525, 625), (-550, -10), (520, -160), (542.5, 475)]

    geom = AGIPD_1MGeometry.from_quad_positions(quad_pos_for_zero_motor_pos)
    geom2 = geom.move_by_motors(motor_pos)
```


> I think the code to read motor positions from saved data (what's currently the `read_motors_from_data()` function) would be a more natural fit as an EXtra component which knows about the source name patterns - this is quite similar to the pattern of components we already have or have planned. It could look something like this:
> 
> ```python
> from extra.components import AGIPDQuadrantMotors
> 
> quad_motors = AGIPDQuadrantMotors(run)
> quad_motors.get_positions(atol=0.01)
> ```

I have no objections. I put all in one place to start discussion. It would be super if it would not need source names. But they depends on the instrument. (yes, we can hard code different sources depending on agipd sources in a run) 

> I don't particularly like the idea of storing motor positions in comments in .geom files - both because it's essentially defining a new ad-hoc file format, and also because if someone modifies the geometry with a tool that ignores the motor positions, they can be out of sync with the module positions they're stored with. But I also don't really have a better idea for how to store these at the moment.

I also don't like. But I think its good enough for now until we find better way. I don't think that is the problem, because if someone modifies geometry with a tool which doesn't know about motors, than we don't know the reason of modification. It's better to discard the motors in this case. Then we need motors only for reference geometry, which instruments supposed to make and keep. Later, I hope, we will store reference geometries in caldb and find a way how to pack the motor positions there.

egorsobolev:
What do you think about mixin implementation?

takluyver:
Are those reference positions you gave real, and accurate enough to be useful? If we can convert absolute motor positions into absolute quadrant positions reliably, I think that opens the door to simpler interfaces, like:

```python
quad_motors = AGIPDQuadrantMotors(run)

geom = AGIPD_1MGeometry.from_motor_positions(
    quad_motors.get_positions(atol=0.01), reference="MID",
)
```

egorsobolev:
The simpler interface would be good.  I gave the numbers from example, but there are real numbers for AGIPD@MID: `[(-542, 660), (-608, -35), (534, -221), (588, 474)]`.

The problem that these numbers is not guaranteed. They can stay the same for any long time, but also may be changed any moment (for example by re-calibrating encoders, and that is not a single option). It means that it is not a good idea to hard code them in library. Thus, simpler interface can look:
```python
quad_motors = AGIPDQuadrantMotors(run)

geom = AGIPD_1MGeometry.from_motor_positions(
    quad_motors.get_positions(atol=0.01),
    reference_quad_pos=[(-542, 660), (-608, -35), (534, -221), (588, 474)]
)
```
Which is actually shortcut replacing just two calls:
```python
geom = AGIPD_1MGeometry.from_quad_positions(reference_quad_pos).move_by_motors(motor_pos)
```

Another problem, that ideal geometry doesn't work for SFX experiments, which are sensitive to the module rotations. That mean we cannot use ideal geometry and need to use refined stored in the file. But we potentially could refine geometry (if we do it with our tools) relative some hard coded origin (move there after refinement using motor positions during the calibration run, instead storing them in the geometry). But then we cannot distinguish geometry files which respect that convention and others.

I would go with explicit way to store and specify motor positions.

takluyver:
Looking at this again:

- As before, I think the code to read motor positions from saved data fits better as an EXtra component rather than here. I can have a go at writing that if you like.
- I don't especially like mixins - I find it makes the inheritance trickier to think about. I think you can do this just as well with an intermediate class: `AGIPD_1MGeometry` -> `DetectorWithMotors` -> `DetectorGeometryBase`. Or even just do it for AGIPD-1M for now, and we can generalise later if we need to. However, this is just stylistic, so feel free to stick with a mixin if you prefer that.
- Reading the motor positions from the file should be in `from_crystfel_geom` rather than `__init__`. The `__init__` method shouldn't assume that the filename it gets is a .geom file, and silencing any errors when reading the file is liable to make for confusing bugs.
- The geometry objects are meant to be basically immutable - public methods to modify the objects always return a new object rather than changing the existing one. As it stands, `set_motor_positions` & `set_motor_axes` break this pattern. I'd like them to either return new objects - (these could be named like `with_motor_positions` or `replace_motor_positions`) or become private methods with a `_` prefix, if they're only meant for use in methods like `.from_crystfel_geom`.
- If the geometry doesn't have motor positions, `move_by_motors` will assume it has motors in the reference/zero positions, which seems wrong & confusing. If we don't know the motor positions to start with, trying to calculate a geometry from new motor positions should throw an error.
  - If we think people will often want to use movements - "motor XYZ was moved 2 mm from where it was just now" - rather than positions, maybe there should be a separate method for that.

egorsobolev:
I created component for motors in European-XFEL/EXtra#148

egorsobolev:
Looking at all suggestions, I decided to refactor it a bit:
- the code to read of motor positions from saved data goes to `EXtra`, see European-XFEL/EXtra#148
- the code to update geometry goes in new class `BaseMotorTracker` and its implementation for specific detectors
- the motor positions stays in the geometry class
- reading the motor positions from the file goes in `from_crystfel_geom`, writing is in `crystfel_fmt.py`
- no one method updates geometry itself, they update `BaseMotorTracker` instead
- the methods, which update geometry or motor position in it, return new geometry instance
- there are two ways to create tracker, with and without reference motor positions
- there are two methods to update geometry `geom_at` and `move_geom_by`; the first one raises an exception if the tracker does not have reference motor positions, the second does not set `motor_positions` to the new geometry

Usage **without** reference motor positions:
```python
    geom = AGIPD_1MGeometry.from_quad_positions(quad_pos)
    tracker = AGIPD_1MMotors(geom)
    geom2 = tracker.move_geom_by(motor_pos)
```

Usage **with** reference motor positions:
```python
    geom = AGIPD_1MGeometry.from_quad_positions(quad_pos)
    tracker = AGIPD_1MMotors.with_reference_positions(geom)
    geom2 = tracker.geom_at(motor_pos)

    tracker = AGIPD_1MMotors.with_reference_positions(geom, ref_motor_pos)
    geom2 = tracker.geom_at(motor_pos)
```

extra_geom/motors.py
```diff
@@ -0,0 +1,214 @@
...
+
+        positions.append(pos)
+
+    return np.array(positions)
+
+
+class BaseMotorTracker:
+    """Detector motor tracker updates geometry according to motor positions.
+    """
+    def __init__(self, ref_geom):
```
takluyver, line 50:
```suggestion
    def __init__(self, ref_geom, ref_motor_positions=None):
```

I think by adding an optional parameter here we can get rid of both `with_reference_positions()` and `set_reference_positions()`. If you want to use new reference positions, you'd just construct a new object:

```python
motor_tracker = AGIPD_1MMotors(motor_tracker, new_ref_positions)
```

extra_geom/motors.py
```diff
@@ -0,0 +1,214 @@
...
+        """
+        ref_motor_positions = np.array(ref_motor_positions, copy=True)
+        if ref_motor_positions.shape != self.motor_position_shape:
+            raise ValueError(f"Expects array{self.motor_position_shape}: "
+                             f"{self.num_groups} groups moving by "
+                             f"{self.num_motors} motor each.")
+
+        self.ref_motor_positions = np.array(ref_motor_positions)
+
+    def set_motor_axes(self, new_motor_axes):
```
takluyver, line 102: Do you see a need for this to be updatable? It looks at a glance like the motor axes will always be hardcoded in the detector-specific subclass.

(It's fine if there is a potential use case, I just like to trim back APIs where possible, and it's generally easier to add things in later than to take them away)

extra_geom/motors.py
```diff
@@ -0,0 +1,214 @@
...
+        """Update geometry according to the motor positions.
+
+        This updates the reference geometry according to the given
+        motor positions with respect the reference motor position.
+
+        Parameters
+        ----------
+        motor_positions: array or list
+            New motor positions as array of the number of movable groups
+            (quadrants) by the number of motor per group.
```
takluyver, line 140: The constant `1e-3` factor in the code implies that these positions are in mm. I'm partly tempted to say 'SI all the way' and put it in metres, but it looks like mm is how motor positions are usually recorded.

The units should definitely be mentioned in the docstring, though - both here and below in `move_geom_by`.

extra_geom/motors.py
```diff
@@ -0,0 +1,214 @@
...
+            raise ValueError(f"Expects array{self.motor_axes_shape}: "
+                             f"{self.num_groups} groups moving by "
+                             f"{self.num_motors} motor each.")
+        self.motor_axes = new_motor_axes
+
+    def geom_at(self, motor_positions):
+        """Update geometry according to the motor positions.
+
+        This updates the reference geometry according to the given
+        motor positions with respect the reference motor position.
```
takluyver, line 134:
```suggestion
        """Get geometry for absolute motor positions

        This returns a new geometry according to the given
        motor positions with respect to the reference motor positions.
```

I think 'update' implies that it's changing the stored reference geometry, which is not the case.

extra_geom/motors.py
```diff
@@ -0,0 +1,214 @@
...
+        motor_diff = motor_positions - self.ref_motor_positions
+        new_geom = self._move_by(1e-3 * motor_diff)
+        new_geom.motor_positions = motor_positions
+        return new_geom
+
+    def move_geom_by(self, motor_diff):
+        """Update geometry according to the changes of motor positions.
+
+        This updates the reference geometry according to the relative changes
+        of motor positions.
```
takluyver, line 167:
```suggestion
        """Get geometry for changes of motor positions.

        This returns a new geometry according to the relative changes
        of motor positions from the reference geometry.
```

takluyver:
I think the structure of this looks good, thanks! I have a few comments about details above.

extra_geom/motors.py
egorsobolev: With two constructors, I'm trying to cover three cases:
1. The tracker without reference positions at all (default constructor). 
2. The tracker with explicitly given reference positions (the second constructor with parameter).
3. The tracker with reference positions which taken from reference geometry instance (the second constructor with default parameter).

In the first case, it's impossible to get geometry by absolute motor positions  (`geom_at` raises an exception)

In the third case, the reference positions must be in the geometry instance. Otherwise an exception is raised.

This is designed to implement the explicit behavior in every case, as your mentioned in the last comment in the previous time. And I like an idea.

In addition, `geom_at` creates the new geometry instance with updated reference position. If the instance is saved later, the reference positions go to the file as well.

In opposite, `move_geom_by` creates the new geometry instance without reference positions.

With one constructor and one parameter with default value is not possible to cover all three cases. It requires one more special value for the parameter or one more parameter. The second constructor looks more natural for me.

`set_reference_position` isolates the code which explicitly set the reference position in the second constructor. In principle, we can get rid of it. But why do not expose if have it anyway?

takluyver: I see, but I wonder if that's confusing if you have a geometry object with motor positions attached and you pass it to `motor_tracker = AGIPD_1MMotors(geom_with_motors)` - this will entirely ignore the motor positions. So I wonder about distinguishing cases 1 & 3 just by whether the geometry has motor information or not.

With regard to my comment from the last round, I agree that if you don't have reference positions, it shouldn't be possible to call `motors.geom_at()`. And I see there's a separate check for that. I'm not sure this needs to be separately checked in a constructor, but I don't feel strongly about this.

egorsobolev: Hmm... Then we don't have an option to discard reference motor position, but we free to ignore the call which is uses them. Ok, let's try to go with it.

egorsobolev: I've pushed changes

extra_geom/motors.py
egorsobolev: Now it is immutable-ish. I get rid of `with_reference_positions`, no clushes

JamesWrigley (commented):
I don't quite understand why we're storing motor positions in the .geom files? I assume the point of saving the metadata is for the sake of reproducibility/provenance; if so, don't we need both the motor positions and the reference geometry?

JamesWrigley:
> The main goal is to store the refined geometry (by SFX data, for example) together with corresponding motor positions. At them moment we use crystfel geometry and sending the geometry files around. Thus, this solution seems natural and transparent.

Sure, but what exactly can you do with that? Do you plan to use them to uniquely identify a particular geometry, or are the positions used by crystfel or something? Sorry if this is a dumb question :sweat_smile: It's just not clear to me how the motor positions are useful without the reference geometry.

egorsobolev:
But the reference geometry is in the same file.

We store refined geometry in file and also store corresponding motor positions there (motor positions at the moment when data for refinement were taken).

Then operators moved motors and ask me to generate new geometry (for example). I read current motor position from Karabo devices, but I need the difference between current positions and positions which correspond to the refined geometry. The last are missed until I store them somewhere (better togehter with refined geometry, which is reference geometry now).

I have implemented writing and reading motor position in the geometry file especially for this particular case. In other cases, this is just a metadata which may be used or not.

egorsobolev:
if there are no more comments... do I merge?

JamesWrigley (approved):
> But the reference geometry is in the same file.
> 
> We store refined geometry in a file (refined by special procedure using the SFX data but not motor changes) and also store corresponding motor positions there (motor positions at the moment when data for refinement were taken).

Ah I see, I misunderstood and thought that the idea was just to store the motor positions with the new geometry.

LGTM!

egorsobolev:
@takluyver and @JamesWrigley, thank you both for review

PR #287, opened 2024-05-22
Prepare to release 1.12

takluyver:
- Add a release note summarising PR #269
- Set up Github actions to publish tags to PyPI, copying the workflow from EXtra-data (I've also set up the PyPI side of this)
- Fix a warning from Sphinx


egorsobolev:
LGTM. I have no expertise in github workflow, here I rely on you.

PR #289, opened 2024-05-22
Transpose default detector axes

egorsobolev:
This fixes the order of default motor axis for AGIPD. The motor device ids follows the notations of `{detector_id}/MOTOR/Q{q}M{m}`, where `M1` corresponds to the vertical axis and `M2` - to the horizontal. It seems natural to keep the same order in the motor position arrays.


takluyver:
LGTM - I haven't checked the details of which motor does what, but I trust you on that.

egorsobolev:
Thanks, @takluyver ! I hope I didn't mix anything up this time.

PR #292, opened 2024-06-18
Fix links to PyFAI documentation

takluyver:
The links with `/master/` in them have got broken. I'll use just https://pyfai.readthedocs.io for the main page, and let it redirect, and `/stable/` for links to specific pages.

I don't think this needs review, so I'll double check the pages and then merge it.


takluyver:
I've checked that these work now, and also opened an issue on pyfai to see if there's a reason the main link defaults to an older version of the docs - https://github.com/silx-kit/pyFAI/issues/2211 .

PR #293, opened 2024-06-18
Switch PyFAI links to use silx.org

takluyver:
@kif tells me the silx.org domain is the preferred location.


takluyver:
Verified the links again.

PR #295, opened 2024-06-25
Increase minimum required Python to 3.8

takluyver:
In line with EXtra-data: https://github.com/European-XFEL/EXtra-data/pull/529


takluyver:
We need to sort out the condat_gridconv dependency first: https://github.com/European-XFEL/condat-gridconv/pull/2

takluyver:
OK, I've updated condat-gridconv, now let's try this again.

PR #303, opened 2024-08-23
Motor tracking for Jungfrau4M

egorsobolev:
%subj%

@takluyver @JamesWrigley

### DIFF
--- a/extra_geom/motors.py
+++ b/extra_geom/motors.py
@@ -212,6 +212,9 @@ class AGIPD_1MMotors(BaseMotorTracker):
 
 
 class JF4MMotors(BaseMotorTracker):
+    # Caution! The assignment of modules to motors and
+    # motors' axes sign are not verified yet.
+
     # groups of modules driven by motors together
     # H1, H2
     groups = [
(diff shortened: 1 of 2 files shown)


takluyver:
Like with EXtra#224, I assume you've tried this out and checked that the results make sense? I don't know the details of JF4M, so I'm happy to trust you on the details of how motors & modules are assigned. The code LGTM, not that there's really any new code here. :slightly_smiling_face:

egorsobolev:
I asked all around and did the best what I can. But we will need to check the axis direction and assignment of the motors to the half next time when we will again refine the geometry. I will cross fingers and merge as is for now.

JamesWrigley (commented):
I think this should be mentioned somewhere in the documentation, but otherwise LGTM!

(though I am absolutely not an expert on geometry)

egorsobolev:
Thanks, @JamesWrigley

JamesWrigley:
Sorry, I actually meant that the class should be in the RST docs :sweat_smile: But it's not urgent, it can wait till you've verified the constants if you like.

egorsobolev:
Hmm... you are right. Looks like, there is no place for it. I think we need a new section

PR #309, opened 2024-09-24
Increase minimum required Python version to 3.9

takluyver:
We already made the same change in EXtra-data.


philsmt:
Seems like a good, LGTM.

philsmt (approved):

PR #330, opened 2025-01-31
Geometry from Karabo

egorsobolev:
This adds script requesting geometry from Karabo

@takluyver @JamesWrigley

### DIFF
--- a/extra_geom/geom_from_karabo.py
+++ b/extra_geom/geom_from_karabo.py
@@ -1,0 +1,93 @@
+# coding: utf-8
+import argparse
+import asyncio
+import base64
+import gzip
+import pickle
+
+from karabo_proxy import AsyncKaraboProxy
+
+FRM_DIMS = {
+    "crystfel": ('frame', 'modno', 'ss', 'fs'),
+    "cheetah": ('frame', 'ss', 'fs'),
+}
+
+presets = {
+    "SPB/AGIPD1M": {
+        "proxy_url": "http://spb-rr-sys-con-gui1:8484",
+        "geometry_device": "SPB_DET_AGIPD1M-1/CAL/GEOMETRY",
+        "z_motor": "SPB_IRU_AGIPD1M/MOTOR/Z_STEPPER",
+        "energy_device": "SPB_XTD2_UND/DOOCS/ENERGY",
+        "clen_min": 0.119,
+    },
+    "SPB/JF4M": {
+        "proxy_url": "http://spb-rr-sys-con-gui1:8484",
+        "geometry_device": "SPB_IRDA_JF4M/CAL/GEOMETRY",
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -36,7 +36,13 @@ def find_version(*parts):
       package_data={
           'extra_geom.tests': ['dssc_geo_june19.h5', 'lpd_mar_18.h5'],
       },
+      entry_points={
+          'console_scripts': [
+              'extra-geom-from-karabo = extra_data.geom_from_karabo:main',
+          ]
+      },
       install_requires=[
+          'Karabo-proxy',
```
takluyver, line 45: Can we make this an optional dependency, using the `extras_require`  section below? If people are installing our tools in their own environment, I try to keep the number of dependencies minimal.

egorsobolev: Hmm, then I we also need to catch exception on import...

takluyver: If you want to, but I think it's also OK to just let the `ImportError` show in that case.

extra_geom/geom_from_karabo.py
```diff
@@ -0,0 +1,93 @@
...
+        "geometry_device": "MID_DET_AGIPD1M-1/CAL/GEOMETRY",
+        "z_motor": "MID_EXP_UPP/MOTOR/T4",
+        "energy_device": "MID_XTD1_UND/DOOCS/ENERGY",
+        "clen_min": 0.0,
+    },
+}
+
+
+def deserialize_geometry(serialized: str):
+    return pickle.loads(gzip.decompress(base64.b64decode(serialized.encode())))
```
takluyver, line 41: I just realised what this is doing.

Unpickling may not work if the version of EXtra-geom installed in Karabo is different from the version locally. Recent versions haven't changed the structure much, but that's not guaranteed for the future, and the failures if you unpickle something from a different version can be very confusing. It also means that anyone calling this is potentially vulnerable to a MITM attack: unpickling is equivalent to `eval()` in security terms.

So I'd rather not do this if we can avoid it. If a major use case is to retrieve a `.geom` file, one option is that the geometry device could expose a slot to create the file, and we call that and write the returned data directly to a file.

D4vidH4mm3r: Relying on pickling in online processing is (as most long-term solutions are) a temporary workaround to the issue of not having settled on a serialization format. I seem to have created [an issue on calng in June of 2023](https://git.xfel.eu/calibration/calng/-/issues/22) to brainstorm serialization.
The issue is of course that we need to decide on a format and that is not so easy. Given that we'll hopefully soon start saving reference geometries in a systematic way, we really need to...

Although we did discuss letting the reference geometry storage be flexible in terms of format. In that case, we can decouple online geometry representation from whatever we put into some database.

Does the CrystFEL `.geom` format adequately allow round-tripping of geometry objects?

PR #338, opened 2025-03-24
Update some bits of outdated info in docs

takluyver:


takluyver:
I don't think this particularly needs review, so I'll merge it now.

PR #343, opened 2025-04-03
Better error messages for unexpected array shapes

takluyver:
Detail the actual & expected dimensions in error messages to make debugging easier.


D4vidH4mm3r:
LGTM! In the issue we looked at today, `expected_data_shape` had been overwritten wrongly, so more investigation would have been needed anyway, but it's nice to see what the assertions would have complained about.

PR #346, opened 2025-04-08
Implementation of _module_coords_to_tile() in JUNGFRAUGeometry

egorsobolev:
this implements messed class method `_module_coords_to_tile` of `JUNGFRAUGeometry`, which is required to transform module to physical coordinates

@takluyver

### DIFF
--- a/extra_geom/detectors.py
+++ b/extra_geom/detectors.py
@@ -2026,6 +2026,13 @@ class JUNGFRAUGeometry(DetectorGeometryBase):
         fs_slice = slice(tile_fs_offset, tile_fs_offset + cls.frag_fs_pixels)
         return ss_slice, fs_slice
 
+    @classmethod
+    def _module_coords_to_tile(cls, slow_scan, fast_scan):
+        row, tile_fs = np.divmod(fast_scan, cls.frag_fs_pixels)
+        column, tile_ss = np.divmod(slow_scan, cls.frag_ss_pixels)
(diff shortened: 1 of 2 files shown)


extra_geom/detectors.py
```diff
@@ -2026,6 +2026,13 @@ def _tile_slice(cls, tileno):
         fs_slice = slice(tile_fs_offset, tile_fs_offset + cls.frag_fs_pixels)
         return ss_slice, fs_slice
 
+    @classmethod
+    def _module_coords_to_tile(cls, slow_scan, fast_scan):
+        row, tile_fs = np.divmod(fast_scan, cls.frag_fs_pixels)
+        column, tile_ss = np.divmod(slow_scan, cls.frag_ss_pixels)
+        tileno = (column * 4 + row).astype(np.int16)
```
takluyver, line 2033:
```suggestion
        column, tile_fs = np.divmod(fast_scan, cls.frag_fs_pixels)
        row, tile_ss = np.divmod(slow_scan, cls.frag_ss_pixels)
        tileno = (row * 4 + column).astype(np.int16)
```

The logic is fine, but the variable names are swapped.

takluyver:
Other than that, LGTM. We can get this in and release a new version today.

extra_geom/detectors.py
egorsobolev: yes, makes sense. thanks!

PR #371, opened 2025-10-30
Add shared header to the docs

JamesWrigley:


JamesWrigley:
(merging as discussed offline)

PR #373, opened 2025-11-13
Allow passing 2D inputs to `JUNGFRAUGeometry`

JamesWrigley:
This saves having to manually insert an extra dimension.


philsmt:
That makes sense, LGTM!

PR #407, opened 2026-04-24
Default vmin & vmax for plotting based on the 2nd & 98th percentiles

takluyver:
Borrowing xarray's [robust plotting option](https://docs.xarray.dev/en/stable/user-guide/plotting.html#robust-plotting), but turning it on by default because detector images often have a few outliers which tends to wash out the default scale.

While I was playing with the plots, I also found the default colour bar unnecessarily large, so I've reduced the fraction matplotlib tries to take for this from 0.15 to 0.07.

Before:

image

After:

image


tmichela:
LGTM

PR #413, opened 2026-05-28
add link to AGIPD paper for more information.

TimBerberich:
Simply adds a link in the AGIPD documentation to a open access paper titled
"The Adaptive Gain Integrating Pixel Detector at the European XFEL"
which contains more details on AGIPD.


tmichela (approved):
LGTM

